PR #6, opened 2025-05-07
Improved test case compilation by compiling each file separately.

Moehre2:
When one test was changed/added all test cases needed to be recompiled. With this change every test file is compiled individually but still linked together to one executable. Now, only the modified file needs to be recompiled when changing/adding a test case.
I checked with `--gtest_list_tests` that all test cases still show up correctly.

(I also performed a bugfix in an include)


Atrisan (approved):

Atrisan (changes requested):
Rebase/merge from main to ensure that the code coverage runs correctly.
Just merged a fix there

Moehre2:
> Rebase/merge from main to ensure that the code coverage runs correctly. Just merged a fix there

Done.

Atrisan (approved):

PR #8, opened 2025-05-08
Json serializer

Atrisan:
File dumping and reading needs to be handled in the docc


Atrisan:
polybench, rodinia, mantevo, and HPCCG for the docc are also tested and work for the CPU.

GPU (kernel node) tests are still pending.

PR #26, opened 2025-05-21
Develop Map node

Atrisan:


src/codegen/dispatchers/map_dispatcher.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+      };
+
+void MapDispatcher::dispatch_node(PrettyPrinter& main_stream, PrettyPrinter& globals_stream,
+                                  PrettyPrinter& library_stream) {
+    LoopSchedule loop_schedule = this->schedule_.loop_schedule(&this->node_);
+    if (loop_schedule == LoopSchedule::VECTORIZATION) {
+        main_stream << "#pragma omp simd" << std::endl;
+    }
+    if (loop_schedule == LoopSchedule::MULTICORE) {
```
lukastruemper, line 20: This is the right way to go, but we have a OMP and Highway Scheduler for For-Loops right now. Let's ignore the loop_schedule set in this PR and we need a second PR to completely re-think the schedule property and dispatchers for the different schedules

src/passes/structured_control_flow/dead_cfg_elimination.cpp
```diff
@@ -133,6 +133,9 @@ bool DeadCFGElimination::run_pass(builder::StructuredSDFGBuilder& builder,
         } else if (auto kernel_stmt = dynamic_cast<structured_control_flow::Kernel*>(curr)) {
             auto& root = kernel_stmt->root();
             queue.push_back(&root);
+        } else if (auto map_stmt = dynamic_cast<structured_control_flow::Map*>(curr)) {
```
lukastruemper, line 136: Maybe add a branch to the is_dead below the for with explicit false.
In a second PR, we could set them as dead iff a) no body and b) the indvar is not used read afterwards

src/passes/structured_control_flow/for2map.cpp
```diff
@@ -0,0 +1,158 @@
...
+                // Add to queue
+                for (size_t j = 0; j < if_else_stmt->size(); j++) {
+                    queue.push_back(&if_else_stmt->at(j).first);
+                }
+            } else if (auto loop_stmt = dynamic_cast<structured_control_flow::While*>(&child)) {
+                auto& root = loop_stmt->root();
+                queue.push_back(&root);
+            } else if (auto for_stmt = dynamic_cast<structured_control_flow::For*>(&child)) {
+                auto num_iterations = this->num_iterations(*for_stmt, analysis_manager);
+                if (symbolic::eq(num_iterations, symbolic::zero())) {
```
lukastruemper, line 123: The num_iterations is not the only criterion. You need a more general: "can_be_applied" and a "num_expressions"

Additional checks of can_be_applied:
- Is the indvar read afterward? (a map does not guarantee that the indvar is the n-1 + n-1 update of the for)
- This can be easily checked with users/happens-before-analysis. Users has a utility function to check whether something is read afterward.
- If it is read afterward, we can apply a trick: Add a transition after the map to the set the indvar to the expected value.

I am not sure if the DataParallelismAnalysis already does that check. We can easily verify by first writing a unit test for it and then check if it produces a wrong result

The can_be_applied needs to check whether the indvar is read afterward. The Map does not guarantee a certain value of the indvar afterward.

src/schedule.cpp
```diff
@@ -31,8 +29,7 @@ const StructuredSDFG& Schedule::sdfg() const { return this->builder_.subject();
 
 analysis::AnalysisManager& Schedule::analysis_manager() { return this->analysis_manager_; };
 
-LoopSchedule Schedule::loop_schedule(
-    const structured_control_flow::ControlFlowNode* loop) const {
+LoopSchedule Schedule::loop_schedule(const structured_control_flow::ControlFlowNode* loop) const {
```
lukastruemper, line 32: Yeah, would be nice to have a parent class in another PR for schedulable CFG elements

src/passes/structured_control_flow/for2map.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    if (!normalizable_update) {
+        return symbolic::zero();
+    }
+
+    auto stride = symbolic::subs(update, index_var, symbolic::zero());
+
+    // Criterion: loop bound must be simple, less than or equal statement (e.g., i < N)
+
+    auto condition = symbolic::rearrange_simple_condition(for_stmt.condition(), index_var);
```
lukastruemper, line 49: Yes, this interesting, as it has been an open question for quite a while:

- LoopBoundNormalizationPass: Converts the simple case of "i != N" into "i < N", if the update is strictly monotonic
- DataParallelismAnalysis: Requires that the update is strictly monotonic and the condition is a simple upper bound "<"
- Map: Basically requires the same as data parallelism analysis, but we have now added a new function which checks it based on new, valuable logic. However, it may be that the data parallelism analysis will fail already much earlier.

So, we should make loop-bound-normalization more powerful probably:
- Instead of doing inequality to "<", we should define something stricter there. This pass runs early.

Then, data parallelism analysis and maps should require the same properties -> API function for both. And this should be automatically fulfilled if loop bound normalization succeeded. Otherwise, it will not be easy to maintain. If someone improves on pass or analysis, it will not automatically improve the others

lukastruemper: I think, I did something recently, which will help here a lot: For the feature extraction, I defined a normal form of conditions. Otherwise, it cannot be encoded as a fixed-dim vector easily.

I required that every condition is in conjunctive-normal-form (CNF), i.e., "Ands of Ors of inequality/equality/symbols". We can probably use the CNF to generalize the re-arrange function

lukastruemper (changes requested):
Very good. It needs to things:

1. Some minor SWA improvements (using the visitor) and adding 1-2 minor things. Easily done first.
2. Loop bounds: I suggest to make the existing LoopBoundNormalization Pass so good that both DataParallelismAnalysis and For2Map accept its output as "applicable for-loops".  To make LoopBoundNormalization powerful we need to refactor the rearrange_condition and the monotonicity functions in symbolic/analysis.cpp. Maybe we can make it very strong by using the conjunctive_normal_form and multi_polynomial analysis functions that I added for the feature extractors recently. Let's discuss after the first step

src/passes/structured_control_flow/dead_cfg_elimination.cpp
Atrisan: I would address this with the Structured Loop nodes introduced in my other branch. That would allow an identical handling of loops and maps and could replace both in the traversal and handling.

I already ported the passes and ~50% of the transformations into this structure.
Will open the PR shortly.

src/passes/structured_control_flow/for2map.cpp
Atrisan: The DataParallelismAnalysis does not check this.
So I will add it by hand as well as a flag to add the transition

Atrisan:
> Very good. It needs to things:
> 
>     1. Some minor SWA improvements (using the visitor) and adding 1-2 minor things. Easily done first.
> 
>     2. Loop bounds: I suggest to make the existing LoopBoundNormalization Pass so good that both DataParallelismAnalysis and For2Map accept its output as "applicable for-loops".  To make LoopBoundNormalization powerful we need to refactor the rearrange_condition and the monotonicity functions in symbolic/analysis.cpp. Maybe we can make it very strong by using the conjunctive_normal_form and multi_polynomial analysis functions that I added for the feature extractors recently. Let's discuss after the first step

Point 1 should be covered.
Regarding the second point, I would address this in an additional PR to not bloat this PR.

lukastruemper (dismissed):
It seems like the data parallelism analysis already checks whether the indvar is used afterward. We need to think about how to handle it later. I.e., For2Map relies on data_parallelism_analysis but there is a solution to it if all bounds are that easy to analyze. Let's merge and normalize the requirements to data parallelism separately

PR #44, opened 2025-05-30
Add Structured Loops

Atrisan:


include/sdfg/structured_control_flow/structured_loop.h
```diff
@@ -0,0 +1,39 @@
...
+   protected:
+    StructuredLoop(size_t element_id, const DebugInfo& debug_info);
+
+   public:
+    virtual ~StructuredLoop() = default;
+
+    StructuredLoop(const StructuredLoop& node) = delete;
+    StructuredLoop& operator=(const StructuredLoop&) = delete;
+
+    virtual const symbolic::Symbol& indvar() const = 0;
```
lukastruemper, line 27: Why shall we require the symbolic description on this type of node? I was thinking of an interface maybe with root as the only virtual func?

lukastruemper (dismissed):

PR #150, opened 2025-07-18
uses serializer for stable clone

lukastruemper:
Serialization and deserialization may be slower than direct copy. However, this increases the code reusage significantly and we have fewer places to maintain library nodes, etc.

### DIFF
--- a/include/sdfg/serializer/json_serializer.h
+++ b/include/sdfg/serializer/json_serializer.h
@@ -21,7 +21,7 @@ class JSONSerializer {
 public:
     JSONSerializer() {}
 
-    nlohmann::json serialize(std::unique_ptr<sdfg::StructuredSDFG>& sdfg);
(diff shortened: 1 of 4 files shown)


NoraHagmeyer:
Could have used my commit for getting rid of the unique pointer on a sdfg :D

According to @Atrisan, the serializer also does not ensure the same ids.

NoraHagmeyer (approved):
Changes make sense, even if they might not ensure a stable clone either

NoraHagmeyer:
> > Could have used my commit for getting rid of the unique pointer on a sdfg :D
> > According to @Atrisan, the serializer also does not ensure the same ids.
> 
> If the serializer does not, then there is a bug. It explicitly recovers the element ids, but it might that we missed some places

Yeah.. still good to only have one place to debug if this problem occurs. Thanks.

PR #154, opened 2025-07-21
Stabilize replay

NoraHagmeyer:


include/sdfg/builder/function_builder.h
```diff
@@ -9,18 +9,19 @@ namespace sdfg {
 namespace builder {
 
 class FunctionBuilder {
-   protected:
```
NoraHagmeyer, line 12: Not sure why this would be happening since we already reformated everything.

PR #166, opened 2025-07-29
Tenstorrent/gemm

ramonwirsch:


src/codegen/dispatchers/block_dispatcher.cpp
```diff
@@ -188,5 +194,51 @@ LibraryNodeDispatcher::LibraryNodeDispatcher(
...
+        if (it == out_edges.end()) {
+            outputs_by_order.push_back(std::nullopt);
+        } else {
+            auto& oedge = *it->second;
+            auto& dst = dynamic_cast<const data_flow::AccessNode&>(oedge.dst());
+            outputs_by_order.push_back(dst.data());
+        }
+    }
+
+    dispatch_code(stream, globals_stream, library_snippet_factory, inputs_by_order, outputs_by_order);
```
Atrisan, line 236: How do the inputs and outputs relate to each other regarding the parameters?
Do you interpret some inputs as outputs?

For C style functions I think this is somewhat confusing, since both lists might overlap, especially since they are ordered.

src/data_flow/library_nodes/math/blas/gemm.cpp
```diff
@@ -528,7 +532,8 @@ void GEMMNodeDispatcher_BLAS::dispatch(codegen::PrettyPrinter& stream) {
         stream << " = " << name << ";" << std::endl;
     }
 
-    if (std::find(gemm_node.inputs().begin(), gemm_node.inputs().end(), "alpha") == gemm_node.inputs().end()) {
+    if (std::find(gemm_node.inputs().begin(), gemm_node.inputs().end(), "alpha") ==
+        gemm_node.inputs().end()) { // TODO obsolute, must be an input!
```
Atrisan, line 536: There is a TODO left...

src/symbolic/symbolic.cpp
```diff
@@ -73,6 +73,10 @@ Expression mul(const Expression& lhs, const Expression& rhs) { return SymEngine:
 Expression div(const Expression& lhs, const Expression& rhs) {
     if (eq(rhs, integer(1))) {
         return lhs;
+    } else if (SymEngine::is_a<SymEngine::Integer>(*lhs) && SymEngine::is_a<SymEngine::Integer>(*rhs)) {
```
Atrisan, line 76: Doesn't the SymEngine do this by itself when using the integer versions, same for the modulo?
Especially, with the extension to simplify, this should be redundant.

Atrisan (changes requested):
The dispatch_code function is still not quite clear to me, especially the ordered inputs and outputs.
A distinction between parameters/arguments and return values is reasonable, but depending on the lib node the output might contain both.

src/data_flow/library_nodes/math/blas/gemm.cpp
ramonwirsch: Yes. That was the intention, although I misspelled obsolete.
What do you put in the code to mark specific points one might later want to update?

Lukas' original modelling left alpha and beta as optional connectors. Where "alpha" was an input-connector only if there was an edge. And if there was no connector then the code would instead use its otherwise irrelevant field.

So it was another hack-implementation for literals / constants in data flow graphs, that solved the same problem as our connector-name as literal abuse with tasklets but did it differently and in a way, where the node-internals need to handle it explicitly.

The new dispatch_code interface was so that handling literals is separate from the codegen, such that we could switch it to literal-nodes without having to update every math-node dispatcher (which is sth. Lukas said we may want to do).

src/codegen/dispatchers/block_dispatcher.cpp
Atrisan: Ok, then I just misunderstood the usage and intention.
The subset resolution can be added once it is needed.

src/symbolic/symbolic.cpp
ramonwirsch: Does it?
For this div impl, we create a func-node, which is essentially just a named placeholder for an operator. It has no implementation. We rely on the language_extension to handle this otherwise not resolvable placeholder.
But we would have never folded the idiv, but would have always kept it symbolically into the final code.

Whereas regular SymEngine operations fold constants during the instantiation. So symbolic::mul(integer(2), integer(3)) is immediatly integer(6) without a need to call simplify.

I wanted to mirror this behavior for our custom integer div operation. Because then it is easy to check if we know the array sizes in for example Gemm statically (if the expression is_a<Integer>()), otherwise we will only know the size at runtime.

I also added this to simplify, such that we would still fold constants in case some other expressions were replaced / resolved to literals after original creation.

src/data_flow/library_nodes/math/blas/gemm.cpp
Atrisan: Thanks for the clarification I just thought that you forgot to handle this.
An issue to be discussed and assigned on Monday might be easier to find again.

Atrisan (approved):

PR #174, opened 2025-08-06
Added flag to infer_type to ignore contiguous memory exception

Moehre2:
As discussed with @Atrisan

### DIFF
--- a/include/sdfg/types/utils.h
+++ b/include/sdfg/types/utils.h
@@ -1,17 +1,11 @@
 #pragma once
 
 #include <cassert>
-#include <map>
 #include <memory>
-#include <sstream>
-#include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


Atrisan:
@lukastruemper this is required for the fallback code gen of einsum-library nodes: https://github.com/daisytuner/sdfglib-einsum/blob/5e7c3c4bb9ac36fe4a95a5e06e0b57b6062d4b9c/src/einsum/einsum_dispatcher.cpp#L42

I think this is preferable over an additional type inference implementation, that needs to be maintained.

lukastruemper (changes requested):
No need for this. It's an integral assumption of the new IR. When you upgraded to a newer sdfglib version, you need to get new SDFGs accordingly, or stay at the old version.

PR #176, opened 2025-08-07
New ConditionElimination Pass

lukastruemper:
Renamed backwardconditionprogation to condition_elimination

Adds a check that the loop variable is not used afterward which is necessary in case the condition is false for iteration 0


rtl/tests/rtl_tests.py
```diff
@@ -78,7 +78,7 @@ def test_instrumentation(event):
     "event",
     [
         pytest.param(""),
-        pytest.param("cuda:::dram__bytes,cuda:::sm__inst_executed_pipe_fma"),
+        # pytest.param("cuda:::dram__bytes,cuda:::sm__inst_executed_pipe_fma"),
```
NoraHagmeyer, line 81: I am guessing this is unrelated to the PR?

NoraHagmeyer (approved):
Only thing is the commented test. But I guess that will be fixed soon.

rtl/tests/rtl_tests.py
lukastruemper: Yes, but the docker image was upgraded in the meantime

PR #181, opened 2025-08-08
Data dependency analysis: Supersede open definitions

lukastruemper:


NoraHagmeyer (dismissed):
Looks fine. Did you use the pre-commit hooks? It seems like clang-format still leaves more leeway than I am used to.

PR #187, opened 2025-08-11
Extends arg-capturing to opaque pointers (on the first level)

NoraHagmeyer:
This is a hot fix to extend arg-capturing to opaque pointers (on the first level). This should be re-done in a clean way once there is an analysis for pointer type inference.

How it works:
- In case that a the argument type is pointer and it does not have a pointee-type, @Atrisan util function is called to infer the type. For that, the sdfg, analysis_manager and container name are necessary. As of right now, this only works if the opaque pointer in on the first nesting level.

How has this been tested:
- Arguments for matmul can be captured
-  Captured arguments can be consumed by the autotuner

Once there is an analysis, maybe we can remove all type inference functionality specific to arg-capturing and base everything on the new analysis.

Question is, if we want this hack in main in the meantime, if I should spend time on improving it now even though everything will change again next week, if we want it broken until we can fix it properly..

### DIFF
--- a/include/sdfg/codegen/code_generator.h
+++ b/include/sdfg/codegen/code_generator.h
@@ -53,7 +53,10 @@ protected:
         int dim_idx,
         const types::IType& type,
         int arg_idx,
-        const analysis::MemAccessRange* range
+        const analysis::MemAccessRange* range,
+        analysis::AnalysisManager& analysis_manager,
+        const StructuredSDFG& sdfg,
+        std::string var_name
     );
 
     bool add_capture_plan(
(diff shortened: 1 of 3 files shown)


ramonwirsch:
If it works for now... go with it.

NoraHagmeyer:
> If it works for now... go with it.

Works for my test case at least.

NoraHagmeyer:
Opened issue to address testing.

NoraHagmeyer:
> If it works for now... go with it.

Was that an approval? Concerning test coverage. We already talked about setting up tests in the future. And I expect to get rid of most of the new code in a week or 2. So, I don't think it's worth writing a unit test for this specific function.

PR #190, opened 2025-08-11
loop analysis - indvar: converts exception into warning

lukastruemper:


NoraHagmeyer:
Does it make sense to do a loop analysis at such an early stage then? Or is the application order of passes the main issue here?

NoraHagmeyer (approved):

PR #191, opened 2025-08-11
Bugfix: StructuredSDFGBuilder::parent did not handle Maps

Moehre2:
The `parent` method of the `StructuredSDFGBuilder` only handels `For` loops. When trying to get the parent of a child of a `Map`, the method returns the root node.
Fixed the code to handle `StructuredLoop`, implicitly handling `Map`.


lukastruemper:
We can merge it but, it is an old API.

Can you please add
```
[[deprecated("use ScopeAnalysis instead)]]
```
to the function.

PR #215, opened 2025-08-28
Revert "Write loopnest index to instrumentation"

lukastruemper:
Reverts daisytuner/sdfglib#213

Loop Nest Indices are an internal name used by the Autotuner. It should be limited to the internals of the Autotuner. The SDFG API is the element ID.

### DIFF
--- a/src/codegen/instrumentation/instrumentation_plan.cpp
+++ b/src/codegen/instrumentation/instrumentation_plan.cpp
@@ -28,9 +28,6 @@ void InstrumentationPlan::begin_instrumentation(const structured_control_flow::C
     stream << metdata_var << ".line_end = " << dbg_info.end_line() << ";" << std::endl;
     stream << metdata_var << ".column_begin = " << dbg_info.start_column() << ";" << std::endl;
     stream << metdata_var << ".column_end = " << dbg_info.end_column() << ";" << std::endl;
-    if (!(this->loopnest_indices_.empty())) {
-        stream << metdata_var << ".loopnest_index = " << this->loopnest_indices_.at(&node) << ";" << std::endl;
-    }
 
     if (this->nodes_.at(&node) == InstrumentationEventType::CPU) {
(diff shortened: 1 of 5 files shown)


src/codegen/instrumentation/instrumentation_plan.cpp
```diff
@@ -65,19 +62,17 @@ std::unique_ptr<InstrumentationPlan> InstrumentationPlan::outermost_loops_plan(S
     auto ols = loop_tree_analysis.outermost_loops();
 
     std::unordered_map<const structured_control_flow::ControlFlowNode*, InstrumentationEventType> nodes;
-    std::unordered_map<const structured_control_flow::ControlFlowNode*, size_t> loopnest_indices;
```
NoraHagmeyer, line 68: This is where the regions are created, aren't they? Where the decision, what will be instrumented and, as of now, uploaded to the database? So, this is where the decision, what will be auto-tuned, is made? We might have different instrumentation strategies in the future, which are not loop nest based, but function based? So, I think it would be beneficial to give all information that are necessary to find this region in the sdfg again?
What I understand is that you are unhappy with the name and use of the the loopnest index to define this. To make this more general, would it make more sense to change the autotuner so that it uses the region name, i.e. sdfg name and outer loop id, to find the region to be tuned in the sdfg and, if necessary, map that to an internal index? This approach could be extended to a more general setting and none of the changes to the instrumentation would be necessary.

NoraHagmeyer:
@Atrisan said that the issue is that the order of outermost loops returned by the loop analysis is unstable. This is a valid reason to move away from using it or to stabilize it. It hasn't created any issue with the "very limited" tests so far and hasn't come up in any of the discussions. If that is the case though, not just this PR needs to b e reverted but auto- as well as transfer tuning need another rework. Potentially, creating a separate indexing scheme if the  one of the loop analysis is unusable.

PR #245, opened 2025-09-24
Lib node offloading

Atrisan:


lukastruemper (changes requested):
Consider using externals instead

lukastruemper (dismissed):

PR #251, opened 2025-09-29
Add support for the ENV var __DAISY_CAPTURE_STRATEGY_DEFAULT to libdaisy_rtl

ramonwirsch:
to control capture strategy.

 * defaults to "once", which is implemented by only capturing the first invocation.
 + "all" to capture all invocations.
 + "never" to skip all capturing, even though the capture-code was injected

 + extended rtl_tests.py with a test case that verifies which invocations are captured by each strategy, including validating the captured data.

### DIFF
--- a/rtl/src/capture.cpp
+++ b/rtl/src/capture.cpp
@@ -10,6 +10,7 @@
 #include <ctime>
 #include <iostream>
 #include <string>
+#include <utility>
 
 #ifndef DEBUG_LOG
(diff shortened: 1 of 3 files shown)


NoraHagmeyer (changes requested):
Thanks for implementing this so fast. Just a few nameing things that may make my life easier in the future as the past has shown that I may have to adapt/modernize/fix some of these parts in the future as one of the only direct users.

PR #630, opened 2026-04-02
Extend Daisyworkflows for PyTorch

Atrisan:
- Reduce harness iterations to safe bench marking time
 - Add additional resnet18 submodels
 - enable warm start for baseline torch backend
 - remove target=none runs


Moehre2 (approved):

Moehre2 (changes requested):
Daisy worflow mlir_torch_models runs in timeout

Moehre2 (approved):

PR #670, opened 2026-04-10
adds is<> utility method to blocks

lukastruemper:


ramonwirsch (dismissed):
"is" alone might be a confusing name as it does not type check the block itself.
Maybe "is_a_single_libnode" would be clearer and leave "is" for future checks on the ControlFlowNodes themselves?

PR #682, opened 2026-04-15
Read only offloads reuse

ramonwirsch:
Offloading can determine that the data remains unchanged and will never even generate a matching D2H. Those could not be removed against further H2D transfers using the same, as-of-yet unchanged input.

 + DataOffloadingNode.is_compatible_with (the part of redundant_with & equal_with that does not look at transfer direction & buffer lifecycle)
 + DataTransferMinimizationAnalysis will now track on-device lifetimes as well
 + on-device matches (H2D->D2H) can be used to reconstruct the missing data and detect read-only on-device data that can be reused as if there was a D2H
 + DataTransferMinimizationPass can delete copy-outs if they have nothing left to do
 ~ also includes fix for DDE: tried to remove arg-containers and did not consider external containers as escape.


Atrisan (approved):
A "Lightweight" Lifetime analysis, I figured this would evolve this way eventually.
But nice.

PR #683, opened 2026-04-15
Tuned ROCM Gemm

Atrisan:
* adds a handtuned gemm implementation for rocm
* extends the GEMM expand for all possible transpose combinations.


opt/include/sdfg/targets/rocm/plugin.h
```diff
@@ -89,19 +90,32 @@ inline void register_rocm_plugin() {
         }
     );
     // GEMM - ROCMBLAS without data transfers
+    // codegen::LibraryNodeDispatcherRegistry::instance().register_library_node_dispatcher(
```
NoraHagmeyer: Can we still leave that implementation in and test it in a unit test if it may be needed again in the future?

This way, it will still be compiled and tested and won't get outdated when other people change the API, tec.

NoraHagmeyer (changes requested):

NoraHagmeyer (approved):

PR #687, opened 2026-04-16
Removes deprecated assumptions analysis functions

lukastruemper:


sdfg/tests/analysis/flop_analysis_test.cpp
```diff
@@ -158,7 +158,7 @@ TEST(FlopAnalysis, Loop) {
     EXPECT_TRUE(symbolic::eq(flop, symbolic::symbol("n")));
 }
 
-TEST(FlopAnalysis, LoopNest) {
+TEST(FlopAnalysis, DISABLED_LoopNest) {
```
lukastruemper, line 161: @Moehre2 This is the only left TODO, but I might want this as a seperate backlog task, because it could go into an "symbolic::overapproximate()" API

PR #689, opened 2026-04-22
Make *LocalStorage* transformations interfaces compatible with Transer Tuning

NoraHagmeyer:
- Adapts the constructor of InLocalStorage, OutLocalStorage, AccumulatorTile, and KernelLocalStorage to take an access node instead of a container
- Adapts GPUTiling to find all reads that are candidates for the KernelLocalStorage Transform instead of container
- Adapts all unit tests and the blocking tests to use the new interface

May change again once the IR is reworked, as access nodes will not exist in this form anymore. Information may be encapsulated in the loop node itself instead.

### DIFF
--- a/opt/include/sdfg/transformations/offloading/gpu_tiling.h
+++ b/opt/include/sdfg/transformations/offloading/gpu_tiling.h
@@ -13,6 +13,7 @@ class GPUTiling : public Transformation {
 
     structured_control_flow::StructuredLoop* inner_loop_ = nullptr;
     structured_control_flow::StructuredLoop* outer_loop_ = nullptr;
+    std::set<std::string> target_accesses_;
 
 public:
     GPUTiling(structured_control_flow::StructuredLoop& loop, size_t size);
--- a/opt/include/sdfg/transformations/offloading/gpu_tiling.h
+++ b/opt/include/sdfg/transformations/offloading/gpu_tiling.h
@@ -13,7 +13,7 @@ class GPUTiling : public Transformation {
 
     structured_control_flow::StructuredLoop* inner_loop_ = nullptr;
     structured_control_flow::StructuredLoop* outer_loop_ = nullptr;
-    std::set<std::string> target_accesses_;
+    std::set<std::string> target_containers_;
 
 public:
     GPUTiling(structured_control_flow::StructuredLoop& loop, size_t size);
--- a/opt/src/transformations/accumulator_tile.cpp
+++ b/opt/src/transformations/accumulator_tile.cpp
@@ -24,8 +24,8 @@
 namespace sdfg {
 namespace transformations {
 
-AccumulatorTile::AccumulatorTile(structured_control_flow::StructuredLoop& loop, std::string container)
-    : loop_(loop), container_(container) {}
+AccumulatorTile::AccumulatorTile(structured_control_flow::StructuredLoop& loop, const data_flow::AccessNode& access_node)
+    : loop_(loop), access_node_(access_node), container_(access_node.data()) {}
 
 std::string AccumulatorTile::name() const { return "AccumulatorTile"; }
 
@@ -409,14 +409,15 @@ void AccumulatorTile::to_json(nlohmann::json& j) const {
     } else {
         throw std::runtime_error("Unsupported loop type for serialization");
     }
-    j["subgraph"] = {{"0", {{"element_id", this->loop_.element_id()}, {"type", loop_type}}}};
+    j["subgraph"] = {
+        {"0", {{"element_id", this->loop_.element_id()}, {"type", loop_type}}},
+        {"1", {{"element_id", this->access_node_.element_id()}, {"type", "access_node"}}}
+    };
     j["transformation_type"] = this->name();
-    j["container"] = container_;
 }
 
 AccumulatorTile AccumulatorTile::from_json(builder::StructuredSDFGBuilder& builder, const nlohmann::json& desc) {
     auto loop_id = desc["subgraph"]["0"]["element_id"].get<size_t>();
-    std::string container = desc["container"].get<std::string>();
     auto element = builder.find_element_by_id(loop_id);
     if (!element) {
         throw InvalidTransformationDescriptionException("Element with ID " + std::to_string(loop_id) + " not found.");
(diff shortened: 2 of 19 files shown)


opt/include/sdfg/transformations/offloading/gpu_tiling.h
```diff
@@ -13,6 +13,7 @@ class GPUTiling : public Transformation {
 
     structured_control_flow::StructuredLoop* inner_loop_ = nullptr;
     structured_control_flow::StructuredLoop* outer_loop_ = nullptr;
+    std::set<std::string> target_accesses_;
```
Atrisan: containers or a set of access_nodes?

opt/src/transformations/offloading/gpu_tiling.cpp
```diff
@@ -1,10 +1,12 @@
 #include "sdfg/transformations/offloading/gpu_tiling.h"
+#include <iostream>
```
Atrisan: cleanup

opt/src/transformations/offloading/gpu_tiling.cpp
```diff
@@ -46,24 +48,32 @@ bool GPUTiling::can_be_applied(builder::StructuredSDFGBuilder& builder, analysis
...
-        KernelLocalStorage kls(*inner_loop, outer_loop->indvar(), container);
+    for (auto& access : read_accesses) {
+        KernelLocalStorage kls(*inner_loop, outer_loop->indvar(), *access);
         if (kls.can_be_applied(builder_local, analysis_manager_local)) {
-            target_containers.insert(container);
+            target_accesses_.insert(access->data());
         }
     }
-    if (target_containers.empty()) {
+    if (target_accesses_.empty()) {
```
Atrisan: We should think about dropping the gpu tiling transformation, since this structure sharing information between the apply and can_be_applied does not allow any scalable application of the pass. (frequent analysis invalidation)

Not a problem for this PR, but something to keep in mind.
Especially for cumulative transformations we need to be very aware of potential invalidations.

opt/src/transformations/offloading/gpu_tiling.cpp
```diff
@@ -100,18 +114,23 @@ void GPUTiling::apply(builder::StructuredSDFGBuilder& builder, analysis::Analysi
     auto& users = analysis_manager.get<analysis::Users>();
     analysis::UsersView users_view_inner_loop(users, inner_loop->root());
 
-    std::set<std::string> target_containers;
+    std::set<data_flow::AccessNode*> target_containers;
```
Atrisan: Why is this a set of access nodes?

opt/src/transformations/offloading/kernel_local_storage.cpp
```diff
@@ -37,9 +37,9 @@ namespace sdfg {
 namespace transformations {
 
 KernelLocalStorage::KernelLocalStorage(
-    structured_control_flow::StructuredLoop& loop, symbolic::Expression offset, const std::string& container
+    structured_control_flow::StructuredLoop& loop, symbolic::Expression offset, const data_flow::AccessNode& access_node
 )
-    : loop_(loop), offset_(offset), container_(container) {};
+    : loop_(loop), offset_(offset), access_node_(access_node), container_{access_node.data()} {};
```
Atrisan, line 42: Why keep both container and access node?

opt/src/transformations/accumulator_tile.cpp
```diff
@@ -24,8 +24,8 @@
 namespace sdfg {
 namespace transformations {
 
-AccumulatorTile::AccumulatorTile(structured_control_flow::StructuredLoop& loop, std::string container)
-    : loop_(loop), container_(container) {}
+AccumulatorTile::AccumulatorTile(structured_control_flow::StructuredLoop& loop, const data_flow::AccessNode& access_node)
+    : loop_(loop), access_node_(access_node), container_(access_node.data()) {}
```
Atrisan, line 28: Same here, do we still require both explicitly?

opt/src/transformations/in_local_storage.cpp
```diff
@@ -24,8 +24,8 @@
 namespace sdfg {
 namespace transformations {
 
-InLocalStorage::InLocalStorage(structured_control_flow::StructuredLoop& loop, std::string container)
-    : loop_(loop), container_(std::move(container)) {}
+InLocalStorage::InLocalStorage(structured_control_flow::StructuredLoop& loop, const data_flow::AccessNode& access_node)
```
Atrisan, line 27: S.O.

opt/src/transformations/out_local_storage.cpp
```diff
@@ -21,8 +21,8 @@
 namespace sdfg {
 namespace transformations {
 
-OutLocalStorage::OutLocalStorage(structured_control_flow::StructuredLoop& loop, std::string container)
-    : loop_(loop), container_(container) {};
+OutLocalStorage::OutLocalStorage(structured_control_flow::StructuredLoop& loop, const data_flow::AccessNode& access_node)
```
Atrisan, line 24: S.O.

Atrisan (changes requested):

opt/src/transformations/offloading/kernel_local_storage.cpp
NoraHagmeyer: Keeping the container_ to not have to change all uses of that since, as you said, the sdfg modeling will probably change drastically soon. Storing the access node mainly for the element ID, so that this can be written to the recording during autotuning.

opt/src/transformations/offloading/gpu_tiling.cpp
NoraHagmeyer: Yes, I had that issue as well. Very frequent invalidations

opt/src/transformations/offloading/gpu_tiling.cpp
NoraHagmeyer: I fused the two loops. Since there is no can_be_applied, there is no stored data anyways. Still, applying transformations while iterating over output of the user analysis is not ideal, but as you said, a separate issue.

opt/include/sdfg/transformations/offloading/gpu_tiling.h
NoraHagmeyer: Changed the name to target_containers_.

Atrisan (approved):

PR #690, opened 2026-04-22
Modular Compile process

ramonwirsch:
+ DoccTarget to allow modularly adding support for snippets that need to be compiled differently (standalone, cross-compiled kernels or just with special compilers)
 + Builder infrastructure to define C/C++ build options with inheritance, overrides and ability to modify by plugins, extensions and targets as needed
 + The redirected handling of snippets based on the extension can contribute link-options when they have been used (the generic solution to highway support)
 + Model to handle any parallism for compile in its own class without having to touch the definitions and compile options
 + DoccPaths extended to resolve default plugin dirs (as used by et-plugin)
 + PrettyPrinter can write directly to another (file) stream instead of a local buffer
 * extended CodeGenerator to allow generating the header first and direct-to-file (its needed for all snippet & main compile, which could be parallel otherwise)
 + Defaults to parallel builds with as many cores as present
 + DOCC_DEBUG is more defined. A list of key-value pairs separated by ; or : Supported options: "dump" dump sdfgs, "build" add debug symbols, "build_threads=x" override thread count. less then 1 is the default auto-selection.
 + CodegenStats is integrated into the new Compiler concept. To reduce the impact of the asynchronicity, times are always measured, but only recored with the CodegenStats if enabled and later, in sequential places. The items recorded now include the sdfg & snippet names to identify slow parts

### DIFF
--- a/c-compile/src/target/docc_target.cpp
+++ b/c-compile/src/target/docc_target.cpp
@@ -1,7 +1,8 @@
 #include "docc/target/docc_target.h"
 #include <filesystem>
+#include <sdfg/targets/highway/codegen/highway_map_dispatcher.h>
 
-#include "docc/compile/file_compiler.h"
+#include "docc/compile/src_file_compiler_builder.h"
 
 namespace docc::target {
 
@@ -18,7 +19,7 @@ static DoccTarget cuda_target = {
         b.add_compile_option("--cuda-gpu-arch=sm_70");
         b.add_compile_option("--cuda-path=/usr/local/cuda");
         b.set_bin_extension("cu");
-        builder.redirect_snippet("cu", b.build());
+        builder.redirect_snippet("cu", std::move(b));
         return true;
     }
 };
@@ -43,7 +44,7 @@ static DoccTarget rocm_target = {
         b.inherit(builder, true);
         b.remove_compile_option("--offload-host-only");
 
-        builder.redirect_snippet("rocm.cpp", b.build());
+        builder.redirect_snippet("rocm.cpp", std::move(b));
 
         return true;
     }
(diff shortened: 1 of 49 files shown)


c-compile/src/target/docc_target.cpp
```diff
@@ -0,0 +1,73 @@
...
+    compile::SrcFileCompilerBuilder highway_builder;
+    highway_builder.inherit(builder, true);
+    highway_builder.contribute_parent_link_options({"-lhwy"});
+
+    builder.redirect_snippet(sdfg::highway::HIGHWAY_SNIPPET_EXT, std::move(highway_builder));
+    return true;
+}
+
+/**
+ * This exists as a workaround for current dependency hell. The builder is only defined in c-compile.
```
NoraHagmeyer: Maybe formulate this a bit more neutral in an open-source repository

opt/include/sdfg/targets/cuda/plugin.h
```diff
@@ -19,117 +20,15 @@
...
-        }
-    );
-
-
-    passes::scheduler::SchedulerRegistry::instance()
-        .register_loop_scheduler<passes::scheduler::CUDAScheduler>(passes::scheduler::CUDAScheduler::target());
+/**
+ * @deprecated use the variant with explicit context
+ */
+inline void register_cuda_plugin() {
```
NoraHagmeyer, line 29: Where is this still used that we need to keep it?

NoraHagmeyer (changes requested):
From what I can tell, looks good, and defines a well-defined interface for targets which are now plugins.

Concerning the compile.. this has become a lot more modular, which is great, but that comes with some new classes of course. Could you draw a small diagram of `codegen_build_pool`, `codegen_compiler`, `src_file_compiler`, `src_file_compiler_builder`, what their task is, and how they interact. This is mainly so that everyone knows about the change, can look it up in a month, and the idea behind your design doesn't get lost.

This also introduces specializations of the debug environment variable, which may be handy for everyone.

opt/include/sdfg/targets/cuda/plugin.h
ramonwirsch: Docc-llvm, sdfglib-auto

It is an exposed interface and we also should, by default strive to remain compatible as that reduces the need to cross module synchronized merges

PR #707, opened 2026-05-08
adds is_tautology api for better analysis of if-else

lukastruemper:


sdfg/src/symbolic/conjunctive_normal_form.cpp
```diff
@@ -169,5 +169,119 @@ CNF conjunctive_normal_form(const Condition cond) {
...
+                pos = true;
+                zero = true;
+                break;
+        }
+        if (neg && zero && pos) return true;
+    }
+    return false;
+}
+
+bool is_tautology(const std::vector<symbolic::Condition>& clause) {
```
ramonwirsch, line 240: I don't quite get what it tests for a tautology?
In the test example it checks that the clauses cover the entire range from one constant point

sdfg/src/structured_control_flow/if_else.cpp
```diff
@@ -43,13 +43,11 @@ bool IfElse::is_complete() const {
...
+    if (symbolic::is_true(condition)) return true;
+
     symbolic::CNF cnf_cond = symbolic::conjunctive_normal_form(condition);
     for (auto& clause : cnf_cond) {
-        auto cond = symbolic::__false__();
-        for (auto& literal : clause) {
-            cond = symbolic::Or(cond, literal);
-        }
-        if (!symbolic::is_true(cond)) {
+        if (!symbolic::is_tautology(clause)) {
```
ramonwirsch, line 50: This will not find all complete cases as it requires an identical constant in all comparisons. For example x < 5, x > 3 is not identified.

sdfg/src/symbolic/conjunctive_normal_form.cpp
lukastruemper: The test currently works by making the OR over all cases (std::vector<Condition>) and checks whether this is always true, to then conclude that the combination of all if-else cases are complete

sdfg/src/structured_control_flow/if_else.cpp
lukastruemper: Yes, this would be even better if we could do this. Right now to goal was to at least be able to reason about ternary if-elses, which symengine does not do itself. More complex would be next step

PR #717, opened 2026-05-19
Add options dict to torch.compile

Moehre2:
Add options dict to torch.compile instead of relying on setting target and category globally with docc.torch.set_backend_options

Also refactored the torch integration tests: By using check_backend and check_compile in almost all tests a lot of redundant code was removed


lukastruemper (approved):
Very nice, even updated the README

PR #718, opened 2026-05-20
Deactivate fastercnn

Moehre2:
The Model Zoo test (nightly) failed every night since it was added. The integration test used a random tensor as input and, therefore, nothing was ever detected by the model. Then, two empty outputs were compared. If the model is run on an actual image the outputs are different.

For now, the nightly run was deactivated and the integration test is skiped until the model is fixed.


NoraHagmeyer (approved):

NoraHagmeyer:
I would see the unit test as an indicator that we don't have any regressions in the compilation part. I will look into the result test at a later point.

PR #719, opened 2026-05-22
Libnode ptr edges

ramonwirsch:
Migrating SDFGs to treat pointers as inputs to libNodes / Calls as scalars.
A pointer will only appear in an output edge if its actually returned from the function (like malloc).
Otherwise its an input edge of the pointer.

Stdlib, Blas and Tensor Matmul nodes were migrated to this new format. Other, currently transitory Tensor Nodes are not yet migrated.


opt/tests/transformations/optimizations/diamond_tiling_test.cpp
```diff
@@ -4,6 +4,7 @@
 #include <fstream>
 #include <memory>
 
+#include "../../../../sdfg/tests/sdfg_debug_dump.h"
```
Moehre2: This does not look right

opt/tests/transformations/optimizations/diamond_tiling_test.cpp
```diff
@@ -180,6 +181,10 @@ TEST(DiamondTilingTest, Jacobi1D) {
     analysis::AnalysisManager am(builder.subject());
     transformations::Recorder recorder;
 
+    dump_sdfg(builder.subject(), "0.before");
+
+    std::cerr << "#####" << std::endl;
```
Moehre2: ?

opt/tests/transformations/einsum2gemm_test.cpp
```diff
@@ -5,6 +5,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "../../../sdfg/tests/sdfg_debug_dump.h"
```
Moehre2: See above

VERSION
```diff
@@ -1 +1 @@
-0.0.8
+0.5.0
```
Moehre2, line 1: Why does the version jump?

Moehre2 (dismissed):

VERSION
ramonwirsch: 0.0.8 was the "docc" version. docc-llvm was on 0.4.0. But the version here (formerly 0.0.8) is now what we use to build the llvm release. So jump forward for docc-compiler (python) and docc-ai (mlir) makes more sense and is not breaking compared to making docc-llvm jump backwards.

And I am assuming we want to release all of them under a common version number as part of the move of docc-llvm here...
So new version needs to be higher than 0.4.0 and 0.0.8.

opt/tests/transformations/optimizations/diamond_tiling_test.cpp
ramonwirsch: Good catch!

PR #720, opened 2026-05-22
Activate Einsum pipeline for Python/PyTorch frontend

Moehre2:
- Added Einsum step to compilation (after which the SDFG is also dumped)
- Generate tasklets and cmath library nodes directly instead of tensor nodes if all tensor types are scalar
- Merged the whole "lifting" part of Einsum nodes into one pass: the EinsumDetectionPass that runs in linear time
- Renamed the EinsumExpand transformation to the EinsumPromotion transformation because the name was confusing with expanding Einsum/library nodes
- Fixed a bug where the lifting failed for multiple input edges of the same access node on a tasklet and added unit tests
- Moved Einsum node source files into tensor folder (but the Einsum node is not a tensor node yet)


NoraHagmeyer (approved):
@ramonwirsch will be happy that you got rid of a pipeline ;)

The unit tests only had to be deleted because you fused the detection into 1 pass, right?

Moehre2:
> @ramonwirsch will be happy that you got rid of a pipeline ;)
> 
> The unit tests only had to be deleted because you fused the detection into 1 pass, right?

Exactly

PR #721, opened 2026-05-22
Make second level tiling possible

NoraHagmeyer:


Moehre2:
Wouldn't it be better to create a `SecondLevelLoopTiling` that inherits from `LoopTiling` instead of having two constructors?

NoraHagmeyer:
> Wouldn't it be better to create a `SecondLevelLoopTiling` that inherits from `LoopTiling` instead of having two constructors?

I changed the code accordingly

Moehre2 (approved):
Looks good :)

PR #724, opened 2026-05-27
Fix bug in DeadDataElimination

Moehre2:
A free node, now, does not have an output connector anymore...


ramonwirsch:
Since no test caught this, can we add a unit test that would have caught it in this PR? Or is the fix urgently needed and we need to schedule the test separately?

ramonwirsch (approved):

ramonwirsch:
There were tests, but using an invalid index in inputs(idx) or outputs(idx) made invalid accesses instead of erroring out.

PR #726, opened 2026-05-29
Fixed LLVM frontend bugs

Moehre2:
- Added c-compile unit tests to github workflow
- Fixed c-compile unit test regarding trailing slashs of paths
- Reactivated -docc-dump-sdfg flag of LLVM frontend to dump intermediate sdfgs to JSON and DOT
- Adapted DOT visualizer to also handle (unstructured) SDFGs
- Fixed a bug where the LLVM frontend created a FreeNode with an output edge
- Fixed a bug where the include files from library nodes where not used in code generation for LLVM frontend


NoraHagmeyer (approved):

NoraHagmeyer:
Very cool!!

For the dump flag.. we could also use the environment variable DOCC_DEBUG=dump to be consistent with the Python frontend. But that is out of scope for now.

PR #742, opened 2026-06-09
Add support for batched gemm nodes in CUDA

Atrisan:
This expands the batched matmul for cuda only to avoid data managment within a loop (batch loop)

### DIFF
--- a/opt/include/sdfg/targets/cuda/blas/batched_gemm.h
+++ b/opt/include/sdfg/targets/cuda/blas/batched_gemm.h
@@ -1,0 +1,46 @@
+#pragma once
+
+#include "sdfg/codegen/dispatchers/block_dispatcher.h"
+#include "sdfg/data_flow/library_nodes/math/blas/batched_gemm_node.h"
+
+namespace sdfg::cuda::blas {
+
+void generate_kernel_batched_gemm(
+    codegen::PrettyPrinter& stream,
+    codegen::LanguageExtension& language_extension,
+    const math::blas::BatchedGEMMNode& node
+);
+
+class BatchedGEMMNodeDispatcher_CUBLASWithTransfers : public codegen::LibraryNodeDispatcher {
+public:
+    BatchedGEMMNodeDispatcher_CUBLASWithTransfers(
+        codegen::LanguageExtension& language_extension,
+        const Function& function,
+        const data_flow::DataFlowGraph& data_flow_graph,
+        const math::blas::BatchedGEMMNode& node
+    );
+
+    void dispatch_code(
+        codegen::PrettyPrinter& stream,
+        codegen::PrettyPrinter& globals_stream,
+        codegen::CodeSnippetFactory& library_snippet_factory
+    ) override;
+};
+
+class BatchedGEMMNodeDispatcher_CUBLASWithoutTransfers : public codegen::LibraryNodeDispatcher {
+public:
+    BatchedGEMMNodeDispatcher_CUBLASWithoutTransfers(
+        codegen::LanguageExtension& language_extension,
+        const Function& function,
+        const data_flow::DataFlowGraph& data_flow_graph,
+        const math::blas::BatchedGEMMNode& node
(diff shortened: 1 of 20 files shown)


opt/include/sdfg/targets/cuda/blas/batched_gemm.h
```diff
@@ -0,0 +1,46 @@
...
+class BatchedGEMMNodeDispatcher_CUBLASWithTransfers : public codegen::LibraryNodeDispatcher {
+public:
+    BatchedGEMMNodeDispatcher_CUBLASWithTransfers(
+        codegen::LanguageExtension& language_extension,
+        const Function& function,
+        const data_flow::DataFlowGraph& data_flow_graph,
+        const math::blas::BatchedGEMMNode& node
+    );
+
+    void dispatch_code(
```
ramonwirsch: deprecated, dispatch_code_with_edges is the successor

sdfg/src/data_flow/library_nodes/math/blas/batched_gemm_node.cpp
```diff
@@ -0,0 +1,524 @@
...
+data_flow::PointerAccessType BatchedGEMMNode::pointer_access_type(int input_idx) const {
+    if (input_idx == 0) { // A
+        auto range = symbolic::mul(batch_count_, stride_a_);
+        return data_flow::PointerAccessMeta::create_read_only(range, true);
+    } else if (input_idx == 1) { // B
+        auto range = symbolic::mul(batch_count_, stride_b_);
+        return data_flow::PointerAccessMeta::create_read_only(range, true);
+    } else if (input_idx == 2) { // C
+        auto range = symbolic::mul(batch_count_, stride_c_);
+        return data_flow::PointerAccessMeta::create_full_write_only(range, true);
```
ramonwirsch: Only true if beta == 0. Otherwise C is also read.
These are optional. you can always return the default (at worst other passes will give up then)

sdfg/src/data_flow/library_nodes/math/blas/batched_gemm_node.cpp
```diff
@@ -0,0 +1,524 @@
...
+
+    auto in_edges = dataflow.in_edges(*this);
+    auto in_edges_it = in_edges.begin();
+
+    data_flow::Memlet* iedge_a = nullptr;
+    data_flow::Memlet* iedge_b = nullptr;
+    data_flow::Memlet* iedge_c = nullptr;
+    data_flow::Memlet* alpha_edge = nullptr;
+    data_flow::Memlet* beta_edge = nullptr;
+    while (in_edges_it != in_edges.end()) {
```
ramonwirsch: we have a dataflow helper. DataFlowGraph.in_edges_by_connector that returns a std::vector<memlet*> that is directly usable (in the order of the connectors as declared in the constructor)

sdfg/src/data_flow/library_nodes/math/blas/batched_gemm_node.cpp
```diff
@@ -0,0 +1,524 @@
...
+        ++in_edges_it;
+    }
+
+    auto* input_node_a = static_cast<data_flow::AccessNode*>(&iedge_a->src());
+    auto* input_node_b = static_cast<data_flow::AccessNode*>(&iedge_b->src());
+    auto* input_node_c = static_cast<data_flow::AccessNode*>(&iedge_c->src());
+    auto* alpha_node = static_cast<data_flow::AccessNode*>(&alpha_edge->src());
+    auto* beta_node = static_cast<data_flow::AccessNode*>(&beta_edge->src());
+
+    if (!input_node_a || dataflow.in_degree(*input_node_a) != 0 || !input_node_b ||
```
ramonwirsch: we have a helper function DataFlowGraph.find_standalone_entry() that returns AccessNode* only if it fullfills the requirements...

ramonwirsch (approved):

PR #743, opened 2026-06-09
Rename interim access nodes for iterative mapfusion calls on nested l…

Atrisan:
…oos to avoid reusing tmp variable introduced by the map fusion.

### DIFF
--- a/opt/src/transformations/map_fusion.cpp
+++ b/opt/src/transformations/map_fusion.cpp
@@ -1,5 +1,6 @@
 #include "sdfg/transformations/map_fusion.h"
 
+#include <iostream>
 #include <isl/ctx.h>
 #include <isl/map.h>
 #include <isl/options.h>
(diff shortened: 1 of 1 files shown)


opt/src/transformations/map_fusion.cpp
```diff
@@ -1,5 +1,6 @@
 #include "sdfg/transformations/map_fusion.h"
 
+#include <iostream>
```
Moehre2: Superfluous include

Moehre2 (changes requested):

opt/src/transformations/map_fusion.cpp
```diff
@@ -773,12 +772,23 @@ void MapFusion::apply(builder::StructuredSDFGBuilder& builder, analysis::Analysi
...
+            std::unordered_map<std::string, std::string> intermediate_renames;
             for (auto& node : first_dataflow.nodes()) {
                 node_mapping[&node] = &builder.copy_node(new_block, node);
                 auto* copied = node_mapping[&node];
                 if (auto* access_node = dynamic_cast<data_flow::AccessNode*>(copied)) {
                     if (access_node->data() == candidate.container) {
                         access_node->data(temp_name);
+                    } else if (first_dataflow.in_degree(node) > 0 && first_dataflow.out_degree(node) > 0) {
+                        // Intermediate access node (e.g. from a prior BlockFusion): clone
+                        // its container so each inlined copy gets its own private scalar
```
lukastruemper, line 786: Missing scalar check. Also not safe because we don't know yet that this is a true local. Would be gone if we had dataflow SSA

lukastruemper (changes requested):

lukastruemper (approved):

PR #744, opened 2026-06-09
Segformer fixes

Moehre2:
- Added partial segformer benchmarks & tests
- Fixed bug in expand_shape and collapse_shape in MLIR frontend that lead to usage of the wrong container
- Fixed bug in depthwise convolution in MLIR frontend where the bias was ignored
- Fixed bug in MapFusion transformation where an access node, accessing the index variable of a map, was not updated when moving to another map
- Visualize assignments in DotVisualizer


opt/src/transformations/map_fusion.cpp
```diff
@@ -779,6 +779,8 @@ void MapFusion::apply(builder::StructuredSDFGBuilder& builder, analysis::Analysi
                 if (auto* access_node = dynamic_cast<data_flow::AccessNode*>(copied)) {
                     if (access_node->data() == candidate.container) {
                         access_node->data(temp_name);
+                    } else if (access_node->data() == first_map_.indvar()->get_name()) {
+                        access_node->data(second_loop_.indvar()->get_name());
```
lukastruemper: MapFusion handles complex relations between first and second map by solving the shift relation between i and j ,e.g.,:

for (i = 1; i <= N; i++) -> for (j = 0; j < N; i++) => i - 1 == j
This is handled in solve_subset: solve_subsets

lukastruemper (changes requested):

lukastruemper (approved):
Good!

PR #745, opened 2026-06-09
Preserve arguments analysis in loop scheduler

NoraHagmeyer:


sdfg/include/sdfg/analysis/analysis.h
```diff
@@ -81,6 +81,24 @@ class AnalysisManager {
         }
     }
 
+    // Preserve only the listed analyses and invalidate all others.
+    // Analyses not present in the cache are unaffected.
+    template<class... Ts>
+    void preserve() {
```
lukastruemper, line 87: This way, we can only preserve one specific analysis at a time

lukastruemper (changes requested):

sdfg/include/sdfg/analysis/analysis.h
lukastruemper: or with one call, but not chain calls... maybe this gets messy

NoraHagmeyer: No, with one call. It basically replaces invalidate_all calls within passes. What would be the use case for chain calls?

lukastruemper (approved):
Agree with current syntax

PR #759, opened 2026-06-13
Extends test suites of local storage and loop transformations

lukastruemper:
- Extends test suite of LoopTiling, ILS, OLS
- Removes cleanup passes from ILS, OLS

### DIFF
--- a/opt/include/sdfg/transformations/in_local_storage.h
+++ b/opt/include/sdfg/transformations/in_local_storage.h
@@ -19,7 +19,7 @@ namespace transformations {
  * within a loop. It copies the accessed subset (tile) into a contiguous buffer
  * before the loop, then redirects all reads to the local buffer.
  *
- * @note The container must be read-only within the loop scope (no writes)
+ * @note The transformation does not check the read-only property.
  * @note All tile extents must resolve to integer constants
  */
 class InLocalStorage : public Transformation {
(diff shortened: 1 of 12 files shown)


opt/include/sdfg/transformations/in_local_storage.h
```diff
@@ -19,7 +19,7 @@ namespace transformations {
  * within a loop. It copies the accessed subset (tile) into a contiguous buffer
  * before the loop, then redirects all reads to the local buffer.
  *
- * @note The container must be read-only within the loop scope (no writes)
+ * @note The transformation does not check the read-only property.
```
NoraHagmeyer: Does that mean we may apply it to variables that get written too and would generate wrong results?

lukastruemper: I am currently trying to make transformations simpler. For ILS and OLS, it would be great to remove the user analysis from the can_be_applied. For ILS, I tried now which would give the responsibility to the user. However, for OLS it does not work because it needs to check whether something is also read. So I am thinking to add those flags is_read and is_write to the MemoryLayoutAnalysis, which both in any case need (and is cheaper). With this, I can add the check back to ILS as well

lukastruemper: So this will be re-introduced..

What is notable though is that we could remove the SequenceFusion and DeadCFGElimination. Additional scopes and deadcfg actually only hurt loop interchange, which in almost any case has happened before a loop becomes the target of local storage transformations, so this would be an efficiency boost for transformations

PR #760, opened 2026-06-13
Add hardware limit checks for gpu

NoraHagmeyer:
Adds gpu hardware limits to avoid silently failing gpu kernel (as in b2_segformer's self attention)


lukastruemper (changes requested):
Would be good to add it to the class' docstring that the transformation is not doing the blocking/tiling on its own, but the scheduler needs to find it. Probably also minimal unit tests

lukastruemper (approved):

PR #762, opened 2026-06-13
Add verifier tests

NoraHagmeyer:
- Adds the verifier tests
- Includes benchmark code via submodule
- Outputs the log/trace for xfail Tests

This is not ideal. In the future, we would probably like to 

- include each external benchmark repository separately
- include NPBench the same way
- unify this over repos (i.e., for the Benchmark repo)
- Run the verifier nightly once we have enough unit test coverage


integration/rodinia_sequential_test.py
Moehre2, line 1: This file was not moved correctly into the llvm/integration/ folder

Moehre2 (changes requested):

Moehre2 (approved):

PR #765, opened 2026-06-15
Faster depthwise convolution

Moehre2:
- Added group attribute to custom conv2D operation and adapted translation
- Set default value for group attribute to 1 (nothing changes)
- Implemented MLIR passes to fuse padding into depthwise convolution by replacing it with custom conv2D operation with groups=channels
- Added different test cases for depthwise convolutions

### DIFF
--- a/mlir/lib/Dialect/Linalg/IR/LinalgCustomOps.cpp
+++ b/mlir/lib/Dialect/Linalg/IR/LinalgCustomOps.cpp
@@ -102,7 +102,7 @@ LogicalResult Conv2DNchwFchwOp::verify() {
     if (inputType.getDimSize(0) != outputType.getDimSize(0)) {
         return this->emitOpError("incompatible N shape");
     }
-    if (inputType.getDimSize(1) != weightsType.getDimSize(1)) {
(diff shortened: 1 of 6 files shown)


mlir/lib/Dialect/Linalg/IR/LinalgCustomOps.cpp
```diff
@@ -102,7 +102,7 @@ LogicalResult Conv2DNchwFchwOp::verify() {
     if (inputType.getDimSize(0) != outputType.getDimSize(0)) {
         return this->emitOpError("incompatible N shape");
     }
-    if (inputType.getDimSize(1) != weightsType.getDimSize(1)) {
+    if (inputType.getDimSize(1) != static_cast<int64_t>(weightsType.getDimSize(1) * this->getGroups())) {
```
NoraHagmeyer, line 105: Is there no better way then a static cast?

NoraHagmeyer (approved):

mlir/lib/Dialect/Linalg/IR/LinalgCustomOps.cpp
Moehre2: As far as I know, this is the best way.

PR #771, opened 2026-06-17
MapFusionByDomain

### LINKED ISSUE #754: Implement support for MapFusion by iteration domain
Allow fusing map stacks (need not end in leaf loops) by matching the iteration domains.

Should be cheaper to execute than existing fusion by memlet and support fusing loops that contain init-reduce and map-reduce patterns.

ramonwirsch:
+ new MapFusion that works simpler by just comparing the iteration domain for each level until there is no more perfectly nested & parallel loop. This way, we can easily fuse [Map] and [Map -> Map] or [Map -> Map] and [Map -> For] by not being forced to fuse all loops in the nest. Its also much easier to check the domain for all the cases where the iteration domains trivially match.

 + new infrastructure to handle updating patterns and continue looking (even inside the updated parts) by updating LoopAnalysis instead of fully invalidating it. Prepwork to make map-fusion scale better.

+ Element.replace(map): version that can replace multiple expressions at once, way more efficient when replacing multiple things on an entiore sub-tree

Closes #754

### DIFF
--- a/opt/src/passes/map_fusion_by_domain_pass.cpp
+++ b/opt/src/passes/map_fusion_by_domain_pass.cpp
@@ -1,4 +1,4 @@
-#include "sdfg/passes/new_map_fusion_pass.h"
+#include "sdfg/passes/map_fusion_by_domain_pass.h"
 
 #include "sdfg/analysis/assumptions_analysis.h"
 #include "sdfg/data_flow/library_nodes/stdlib/malloc.h"
@@ -8,7 +8,7 @@
 
 namespace sdfg::passes {
 
-FusionLoopCandidate* NewMapFusionPass::State::get_next_level_map_stack(FusionLoopCandidate& current) {
+FusionLoopCandidate* MapFusionByDomainPass::State::get_next_level_map_stack(FusionLoopCandidate& current) {
     auto& children = loop_analysis->children(current.loop);
     if (children.empty()) {
         return nullptr;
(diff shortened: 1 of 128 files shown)


opt/src/passes/map_fusion_by_domain_pass.cpp
```diff
@@ -0,0 +1,693 @@
...
+    auto& arguments_analysis = analysis_manager.get<analysis::ArgumentsAnalysis>();
+
+    for (auto* control_flow_node : state.loop_analysis->loops()) {
+        if (auto* map = dynamic_cast<Map*>(control_flow_node)) {
+            auto& indvar = map->indvar();
+            auto& assumpts = assumption_analysis.get(map->root(), true);
+            auto* indvar_boundaries = find_indvar_boundaries(indvar, assumpts);
+
+            if (indvar_boundaries && !indvar_boundaries->tight_lower_bound().is_null() &&
+                !indvar_boundaries->tight_upper_bound().is_null() && !indvar_boundaries->map().is_null()) {
```
lukastruemper, line 240: Should be implied anyway by a map, but does not harm

opt/src/passes/map_fusion_by_domain_pass.cpp
```diff
@@ -0,0 +1,693 @@
...
+        // replace on it, then flatten it into the existing
+        append_root = &state_.builder.add_sequence_before(target_root, target_root.at(0).first, {}, {});
+    }
+
+    std::optional<std::unordered_map<const ControlFlowNode*, const ControlFlowNode*>> copy_mapping;
+    if (can_remove_original) {
+        state_.builder.move_children(first_current->loop->root(), *append_root);
+    } else {
+        deepcopy::StructuredSDFGDeepCopy copier(state_.builder, *append_root, first_current->loop->root());
+        copy_mapping = copier.insert();
```
lukastruemper, line 311: Keep in mind that we reuse the indvars of nested loops when deep-copying, which adds WAW dependencies of indvars, but something to be solved with LoopSSA

lukastruemper (dismissed):
Strong PR, nice. Have some comments to keep in mind

opt/src/passes/map_fusion_by_domain_pass.cpp
ramonwirsch: Yes. This might become a problem in the future. Currently we only move loop contents. As long as loops themselves use unique containers on creation, they still should after those moves.

ramonwirsch:
Rocm segformer tests fail due to LoopDistribute splitting up local var write and local var read, creating an incorrect SDFG.

lukastruemper (approved):
As discussed, current limitations caused by downstream passes need to be tracked in new issues.

PR #802, opened 2026-06-22
adds reduce as new structured loop type

lukastruemper:


python/benchmarks/npbench/deep_learning/mlp/test_mlp.py
```diff
@@ -60,19 +60,21 @@ def test_mlp(target):
     if target == "none":
         verifier = SDFGVerification(
             verification={
-                "SEQUENTIAL": 18,
+                "SEQUENTIAL": 20,
                 "FOR": 20,
                 "MAP": 18,
+                "REDUCE": 2,
```
ramonwirsch: How can it be that reduce increases without for decreasing, but sequential does?
Are we counting every reduce still as for? And how does that change the sequential count?
Are we counting the root sequences for non-reduce structured loops but not for reduce?

sdfg/src/serializer/json_serializer.cpp
```diff
@@ -224,10 +223,14 @@ void JSONSerializer::block_to_json(nlohmann::json& j, const structured_control_f
     }
 }
 
-void JSONSerializer::for_to_json(nlohmann::json& j, const structured_control_flow::For& for_node) {
+void JSONSerializer::for_to_json(nlohmann::json& j, const structured_control_flow::StructuredLoop& for_node) {
```
ramonwirsch: Can we rename this to structured_loop_to_json to reduce confusion?

ramonwirsch (changes requested):

PR #806, opened 2026-06-23
Map Collapse for Multiple targets in a neste sequence

Atrisan:
Extend the map collapse to allow collapsing map nests with multiple elements in a sequence, similar to a cuda kernel.

* using the largest loop as a maximum inner iteration count
* Adding guards for smaller or non-loop elements


lukastruemper (approved):
Nice work

PR #808, opened 2026-06-23
enable batched gemm and memcpy on AMD GPUs analogous to the existing CUDA implementation

Atrisan:


Moehre2 (dismissed):

ramonwirsch (approved):
Mid-term we have to find new ways to have less code dupliated between cuda & rocm...

PR #810, opened 2026-06-24
Add DOCC_REUSE_SOURCES envar to recompile the shared library from locally edited sources

Atrisan:


mlir/docc/torch/torch_program.py
```diff
@@ -280,6 +292,26 @@ def compile(
             self._device_resident = sdfg.metadata("device_resident") == "1"
             backend = sdfg.metadata("device_backend")
             self._device_backend = backend or None
+        elif docc_reuse_sources:
+            main_file = f"{output_folder}/__docc_GraphModule.cpp"
```
Moehre2: Hardcoding the name "__docc_GraphModule" is not safe. This is the name of the SDFG. Simply put the name property of the SDFG here.

Moehre2 (changes requested):

mlir/docc/torch/torch_program.py
```diff
@@ -235,9 +235,24 @@ def compile(
         # Reuse already built binaries
         docc_reuse_binaries = os.environ.get("DOCC_REUSE_BINARIES")
 
+        # Reuse already generated sources (recompile without regenerating them).
+        # Unlike binary reuse this still runs the full pipeline, but the build
+        # step recompiles the existing source files instead of overwriting them.
+        docc_reuse_sources = os.environ.get("DOCC_REUSE_SOURCES")
+
+        if docc_reuse_binaries or docc_reuse_sources:
+            os.environ["DOCC_DEBUG"] = "dump"
```
Moehre2: This is not correct. DOCC_DEBUG can take multiple arguments. E.g., `DOCC_DEBUG=dump;build`. If DOCC_DEBUG is already set like this and I set DOCC_REUSE_SOURCES, it will overwrite DOCC_DEBUG and remove the `build` part.

Moehre2 (changes requested):

Moehre2 (approved):

PR #811, opened 2026-06-24
Add a pass to make loop indvars unique again

Atrisan:


lukastruemper (changes requested):
Either traverse the SDFG (maybe there's a helper from other new analysis) to check whether the indvar was read after the loop and before the next write happens. There may be actual dependencies (rodinia/llvm-test-suite have those benchmarks).

ramonwirsch (approved):

PR #812, opened 2026-06-25
Define Buffer reuse pass for offloaded containers

Atrisan:
* Creates a simplified kernel dependency graph with data offload node
* Uses Dominace to find candidates for elimination
* Heuristic, reuse biggest buffers first
* Can distunguish between pure control flow ordering in sequences (default) and concurrent data flow dependencies.


opt/src/passes/offloading/device_buffer_reuse_pass.cpp
```diff
@@ -0,0 +1,435 @@
...
+            continue;
+        }
+
+        // The colouring order is only a heuristic: with symbolic sizes `Gt` is not a total order
+        // (e.g. S vs. T compares false both ways), so members.front() is not guaranteed to be the
+        // largest. It still serves as the shared allocation's name, but the surviving allocation
+        // must be sized to the symbolic maximum over all members.
+        const std::string representative = candidates[members.front()].container;
+        symbolic::Expression max_size = candidates[members.front()].size;
+        for (size_t m : members) {
```
ramonwirsch, line 379: probably more efficient to use SymEngine::max, which gets a vector of elements instead of building a unbalanced tree of binary ops that will get simplified to the former

ramonwirsch (dismissed):

PR #813, opened 2026-06-26
Add Templates and OpenSource Structure

Atrisan:


.github/CODEOWNERS
```diff
@@ -0,0 +1,27 @@
...
+# Each line maps a path to one or more owners (GitHub usernames or teams).
+# All entries below are currently UNASSIGNED — add owners after the path,
+# e.g. "/python/  @your-username @your-team".
+#
+# Docs: https://docs.github.com/articles/about-code-owners
+
+# Default owner for everything (unassigned)
+# *
+
+# Top-level directories (unassigned)
```
NoraHagmeyer, line 24: I have worked with code owners for 5 years, and my experience is that this becomes a bottleneck, i.e. we had problems that some parts of the code were only owned by a team of 4 people. So, we really cannot have individual people owning code. This will block progress.

My suggestions: Create teams. Add people to teams. code is owned by a team. Each team should have at least 2 people in it. As we grow, there cannot be one part of the code that is only maintained by one person.

.github/CODEOWNERS
```diff
@@ -0,0 +1,27 @@
...
+/arg-capture-io/ @NoraHagmeyer
```
NoraHagmeyer: Ramon and me can own this

.github/CODEOWNERS
```diff
@@ -0,0 +1,27 @@
...
+/c-compile/ @ramonwirsch
```
NoraHagmeyer: Ramon cannot be the only person who has an overview of how the compile step in our compiler works. If nobody knows that yet (I reviewed the PR and I already debugged a problem due to it), someone from the compiler team needs to learn.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Type of Change
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Refactor / internal change (no functional change)
+- [ ] Documentation update
+
+## Quality Checklist
+*Please ensure the following are completed before requesting review:*
```
NoraHagmeyer, line 21: I would add something like [ ] These changes preserve backward compatibility. If that is not ticked, everyone should get mentioned. If that is ticked and not true, the person did not think about the dependencies/ impact on other parts of the code.

NoraHagmeyer (changes requested):
I like the templates and the contributing guidlelines. I think this will make things much easier for new contributors.

I have some strong opinions on the code owners based on 5 years of experience developing in a code base with 25+ active maintainers.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,39 @@
...
+A clear and concise description of what you expected to happen.
+
+## Screenshots / Video
+If applicable, add screenshots or screen recordings to help explain your problem.
+
+## Environment Information
+Please complete the following information:
+- **OS/OS Version:** [e.g. Ubuntu 24.04, macOS 13.4]
+- **Python Version:** [e.g. 3.11.4]
+- **LLVM Version:** [e.g. 19.1.4, commit hash `a1b2c3d`]
```
ramonwirsch: We may want to ask for the source of the package (distro, llvm repos, etc.) for package-installed once or commit-hash for ones built from source.
Especially the CMake setup may differ by maintainer. And hashes might not be all to relevant or easy to get for distro-packaged versions.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,39 @@
...
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+## Expected Behavior
+A clear and concise description of what you expected to happen.
+
+## Screenshots / Video
+If applicable, add screenshots or screen recordings to help explain your problem.
```
ramonwirsch: "Screen Recordings" on a compiler issue as long as we don't have a UI might incentivize more stupid submissions...

.github/ISSUE_TEMPLATE/enhancement_request.md
```diff
@@ -0,0 +1,31 @@
...
+## Pre-flight Checklist
+*Please ensure you have completed the following tasks before submitting this issue:*
+- [ ] I have searched existing issues to ensure this feature hasn't been requested before or is not already being worked on.
+- [ ] This request targets a single focused feature or improvement (not a massive multi-part epic).
+
+## Is your feature request related to a problem?
+A clear and concise description of what the problem is. Ex. "I'm always frustrated when..."
+
+## Proposed Solution
+A clear and concise description of what you want to happen. What should the UX/UI look like?
```
ramonwirsch: Does not have UI. 
Result? Code? Graph? not sure. But UI may confuse

.github/PULL_REQUEST_TEMPLATE.md
ramonwirsch: We could make this more explicit & detailed maybe like this:

IR:
  

This should make it easier for people to understand how they could have affected compatibility.

CONTRIBUTING.md
```diff
@@ -0,0 +1,100 @@
...
+2. Make sure submodules are initialized:
+   ```bash
+   git submodule update --init --recursive
+   ```
+3. Install the prerequisites for the component you want to work on. The project
+   is organized into modules, each with its own `README.md`:
+   - `sdfg/` — core SDFG intermediate representation, passes, and analyses.
+   - `python/` — Python (JIT) frontend and bindings.
+   - `mlir/` — MLIR dialect and PyTorch frontend.
+   - `opt/` — target code generation (Generic, OpenMP, CUDA, ROCm).
```
ramonwirsch: This does not reflect our current goals.
We need to migrate to have target specific libs, even for default-included targets due the dependencies on of our DoccTarget mechanisms.
So opt should more contain optional optimization passes and logic. This may include code reused across rocm and cuda targets for example, but the TargetDefinition needs to be in a separate library and so its not that useful to put non-generic & non-reusable target specific code here.

Also, interfaces that plugins & targets depend on currently need to live in sdfg/ (so LoopScheduler, which is provided by targets / plugins)

CONTRIBUTING.md
```diff
@@ -0,0 +1,100 @@
...
+   - `rtl/` — runtime libraries for instrumentation.
```
ramonwirsch, line 40: - targets/... -- new targets should live here
 - c-compile -- code to emit binaries, source code or lower it to another compiler-IR (such as LLVM-IR)

ramonwirsch (dismissed):

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,39 @@
...
+name: "Bug Report"
+about: Create a report to help us improve and fix bugs.
+title: "[BUG] <Short, descriptive title>"
+labels: bug
+assignees: ''
+---
+
+## Pre-flight Checklist
+*Please ensure you have completed the following tasks before submitting this issue:*
+- [ ] I have verified this bug exists on the latest version of the master/main branch.
```
Moehre2: Does any of our repositories use a "master" branch? I would simply write main branch.

.github/CODEOWNERS
Moehre2: I strongly agree

Moehre2 (changes requested):

NoraHagmeyer (dismissed):

ramonwirsch (dismissed):

.github/CODEOWNERS
```diff
@@ -0,0 +1,38 @@
...
+
+# Top-level directories (unassigned)
+/arg-capture-io/ @NoraHagmeyer @daisytuner/tuning-interface
+/c-compile/ @ramonwirsch @daisytuner/compiler
+/cmake/ @ramonwirsch @daisytuner/compiler
+/examples/ @Atrisan @daisytuner/compiler
+/llvm/ @Moehre2 @daisytuner/compiler
+/mlir/ @Moehre2 @daisytuner/compiler
+/opt/ @Atrisan @daisytuner/compiler
+/python/ @Moehre2 @daisytuner/compiler
```
Moehre2, line 32: Add `/pytorch/ @Moehre2 @daisytuner/compiler` line here :)

Moehre2 (changes requested):

NoraHagmeyer (approved):

Moehre2 (approved):

PR #814, opened 2026-06-26
Adds GPU reduce dispatchers

lukastruemper:


opt/src/targets/cuda/cuda_reduce_dispatcher.cpp
```diff
@@ -0,0 +1,543 @@
...
+            // 2^(width-1) - 1
+            if (width == 32) return "2147483647";
+            if (width == 64) return "9223372036854775807LL";
+        }
+    } else { // Max -> identity is the type minimum
+        if (is_unsigned) {
+            return "0" + suffix;
+        }
+        if (width == 32) return "(-2147483647 - 1)";
+        if (width == 64) return "(-9223372036854775807LL - 1)";
```
Atrisan: There is \<climits\> which has the integer min and max values as predefined constants.

opt/src/targets/cuda/cuda_reduce_dispatcher.cpp
```diff
@@ -0,0 +1,543 @@
...
+    }
+    if (op == ReductionOperation::Mul) {
+        return "1";
+    }
+
+    // Min / Max: need the most-extreme element so the first real value wins.
+    const bool floating = types::is_floating_point(prim);
+    if (floating) {
+        // INFINITY / -INFINITY are provided by CUDA's math headers.
+        return op == ReductionOperation::Min ? "INFINITY" : "-INFINITY";
```
Atrisan: Unsafe, if we ever want to extend the support for reductions

opt/src/targets/cuda/cuda_reduce_dispatcher.cpp
```diff
@@ -0,0 +1,543 @@
...
+        if (is_unsigned) {
+            // 2^width - 1
+            if (width == 32) return "4294967295U";
+            if (width == 64) return "18446744073709551615ULL";
+        } else {
+            // 2^(width-1) - 1
+            if (width == 32) return "2147483647";
+            if (width == 64) return "9223372036854775807LL";
+        }
+    } else { // Max -> identity is the type minimum
```
Atrisan: Unsafe, if we ever want to extend the support for reductions

opt/src/targets/cuda/cuda_reduce_dispatcher.cpp
```diff
@@ -0,0 +1,543 @@
...
+    library_stream << "{" << std::endl;
+    library_stream.setIndent(library_stream.indent() + 4);
+
+    this->dispatch_reduce_core(
+        library_snippet_factory, library_stream, CUDADimension::X, /*declare_scope_variables=*/true, scope_variables
+    );
+
+    library_stream.setIndent(library_stream.indent() - 4);
+    library_stream << "}" << std::endl;
+};
```
Atrisan: Now that we have this boilerplate more than once (Map and Reduce) consider moving it into its own function to avoid replication

opt/src/targets/rocm/rocm_reduce_dispatcher.cpp
```diff
@@ -0,0 +1,543 @@
+#include "sdfg/targets/rocm/rocm_reduce_dispatcher.h"
```
Atrisan, line 1: A lot of replication from the CUDA equivalent, especially for functions that are not device specific.

Atrisan (changes requested):

ramonwirsch (dismissed):

Atrisan (approved):

PR #830, opened 2026-07-03
Adapted LLVM test suite to main

Moehre2:
No additionally running tests :(

Deactivated siod because it does not compile with leak sanitizer


NoraHagmeyer:
We lost a test? :(

NoraHagmeyer (approved):

PR #843, opened 2026-07-06
adds type_ids for Element classes and a human-readable element_type

lukastruemper:


opt/src/targets/cuda/cuda_data_offloading_node.cpp
```diff
@@ -176,15 +176,15 @@ codegen::InstrumentationInfo CUDADataOffloadingNodeDispatcher::instrumentation_i
     if (cuda_node.is_d2h()) {
         return codegen::InstrumentationInfo(
             node_.element_id(),
-            codegen::ElementType_D2HTransfer,
```
ramonwirsch, line 179: This is a breaking change and we do not have a use for this info being included in tracing.
And if full SDFG info is needed, we can lookup the element ID in an SDFG dump.

opt/src/targets/rocm/rocm_data_offloading_node.cpp
```diff
@@ -174,15 +174,15 @@ codegen::InstrumentationInfo ROCMDataOffloadingNodeDispatcher::instrumentation_i
     if (rocm_node.is_d2h()) {
         return codegen::InstrumentationInfo(
             node_.element_id(),
-            codegen::ElementType_D2HTransfer,
```
ramonwirsch, line 177: Same as above

ramonwirsch (changes requested):
We want to keep specifically D2H and H2D types in a tracing-specific way, such that the backend does not need to be aware of all OffloadingNode impls to keep the current functionality.

tutorial/printf_target/src/printf_data_offloading_node.cpp
```diff
@@ -158,15 +158,15 @@ codegen::InstrumentationInfo PrintfDataOffloadingNodeDispatcher::instrumentation
     if (printf_node.is_d2h()) {
         return codegen::InstrumentationInfo(
             node_.element_id(),
-            codegen::ElementType_D2HTransfer,
+            std::string(printf_node.element_type()) + ":::" + printf_node.code().value(),
```
ramonwirsch, line 161: same here

targets/tenstorrent/src/tenstorrent/tenstorrent_offloading_node.cpp
```diff
@@ -337,15 +337,15 @@ codegen::InstrumentationInfo TTDataOffloadingNodeDispatcher::instrumentation_inf
     if (tt_node.is_d2h()) {
         return codegen::InstrumentationInfo(
             node_.element_id(),
-            codegen::ElementType_D2HTransfer,
```
ramonwirsch, line 340: same here

PR #845, opened 2026-07-07
Stacktraces for python and mlir bindings

ramonwirsch:
Uses Boost::stacktrace, statically linked into python and mlir bindings to inject stacktraced at throw-time into exceptions and tries to sniff all exceptions passing through pybind (so uncaught exceptions) and dump their stacktrace to stderr.

Only on linux systems do we enable building and linking of Boost::stacktrace_from_exception, which does the actual injecting, as MacOS env is currently unsupported.

Note: as the python bindings are only loaded at runtime from a python script, it loads too late to inject stacktraces into all the other C++ libraries that have already been loaded. To also inject those, you would need to build boost 1.90 dynamic libs locally and LD_PRELOAD them.


Moehre2 (approved):
Nice :)

PR #851, opened 2026-07-15
Add map fusion handling of analyses without invalidation

NoraHagmeyer:


opt/src/passes/normalization/map_fusion.cpp
```diff
@@ -19,11 +20,22 @@ bool MapFusion::accept(structured_control_flow::Sequence& node) {
         return applied;
     }
 
+    // LoopAnalysis is cheap to re-run and must be re-run whenever the loop structure changes.
```
ramonwirsch, line 23: new LoopAnalysis does not need to be re-run it already can be updated. Just fyi.

opt/src/passes/normalization/map_fusion.cpp
```diff
@@ -41,7 +61,26 @@ bool MapFusion::accept(structured_control_flow::Sequence& node) {
...
-                transformation.apply(builder_, analysis_manager_);
+
+                // Apply without letting the transformation invalidate the manager;
+                // we selectively invalidate below based on the fusion direction.
+                transformation.apply_without_invalidate(builder_, analysis_manager_);
+
+                if (transformation.was_producer_into_consumer()) {
+                    // ProducerIntoConsumer only prepends Block nodes to second's body.
+                    // The loop structure of every other node in this sequence is unchanged,
+                    // so AssumptionsAnalysis and ArgumentsAnalysis stay valid.
```
ramonwirsch, line 72: Can it not happen, that the fusion fuses further memory accesses into the 2nd loop that would add arguments?
i.e.
map { a[i] = in[i] }
map { b[i] = a[i] }
-> map { b[i] = in[i] } with (currently) a no longer being an argument and in being a new one.

NoraHagmeyer: I only know that all tests pass. But that's why I never wanted it merged but cleanly done. With the goal of bringing down the compile time though.

opt/src/passes/normalization/map_fusion.cpp
NoraHagmeyer: Ah, I did not want to open the whole "updating an analysis" bucket

PR #853, opened 2026-07-15
Make LoopSchedulers available from DoccTarget

ramonwirsch:
+ Each target can declare its loop schedulers (also multiple) when it gets applied
 * global scheduler registry no longer needed (in theory)
 * Switched LoopSchedulingPass over to using LoopScheduler ptrs directly instead of strings it will internally resolve to LoopSchedulers using global state.
 * Also extracted some utilities from CudaOffloading and TenstorrentOffloading for reuse in further target / offloading code
 * LLVM Frontend was also migrated to updated LoopSchedulingPass, but since target and context have not been deeply integrated yet, its implementation lags behind.


sdfg/include/sdfg/plugins/plugins.h
```diff
@@ -15,21 +15,63 @@
...
+/**
+ * TODO this class is broken. We need an interface that exposes the references on methods. So that there can be a future
+ * implementation that actually owns the different registries as local state. And this legacy adapter can exist that
+ * just redirects to problematic broken state. The methods exist, so we can migrate users over to the methods, before
+ * stripping the implementation-specific parts into subclasses.
+ *
+ * Also, SchedulerRegistry currently lives not in this library, so poses another API hazard, as you could compile this
+ * as part of the library it is in!
+ */
 struct Context {
```
Atrisan, line 27: Just make it a class instead and avoid all those deprecations

opt/include/sdfg/passes/scheduler/scheduler_registry.h
```diff
@@ -7,6 +7,9 @@ namespace sdfg {
 namespace passes {
 namespace scheduler {
 
+/**
+ * @deprecated Global state and superfluous, do not use
```
Atrisan, line 11: should be used in exactly 2 places (llvm frontend and python fronend) just update them instead of deprecating.

Atrisan (changes requested):

sdfg/include/sdfg/plugins/plugins.h
ramonwirsch: Deprecations are a requirement to not break cross-repo code. Which sdfglib-auto and external plugins for example are.

That is simply the nature of referencing one lib from a different repo. You need a transitory commit where both exists, until the external user has been migrated away. But it also cannot migrate away if the new code does not exist yet.

Atrisan (approved):

PR #856, opened 2026-07-17
Migrating Sequence Transitions into their own AssignmentBlock ControlFlowNode

ramonwirsch:
+ AssignmentBlock is a 1:1 Replacement for Transitions. It is now basically optional and no longer tied to a Sequence.
 + Sequence.at only returns the node, no longer a pair of node & transition
 * a few passes now need to create AssignmentBlocks to put before or after if/loop/sequence assignments. These need to be careful to not infinitely create new blocks in case the pass is called from a pipeline, but also needs to validate that there is no collision with assignments that may be present in a coincendantally pre-existing AssignmentBlock in the same place (then they need to create a new one in between).
 * if those passes definitely won't run in a pipeline anymore, this can be simplified to always create a fresh block


ramonwirsch:
RPC Tests fail because the slightly different SDFG structure may not match. But even more importantly, the live server runs not on a docc version matching this branch. i.e. it does not support AssignmentBlocks at all, causing 500 errors from the backend.

This should solve the cpp-tests-linux and Transfertuning failures.

sdfg/include/sdfg/structured_control_flow/block.h
```diff
@@ -2,6 +2,7 @@
 
 #include <memory>
 
+#include "sdfg/control_flow/interstate_edge.h"
```
lukastruemper, line 5: ?

opt/src/passes/scheduler/loop_scheduling_pass.cpp
```diff
@@ -17,9 +17,6 @@ bool LoopSchedulingPass::run_pass_target(
     scheduler.set_report(report_);
     scheduler.set_recorder(recorder_);
 
-    UniqueLoopIndvars unique_indvar_pass;
-    unique_indvar_pass.run_pass(builder, analysis_manager);
```
lukastruemper, line 21: Can we simply remove it?

lukastruemper (changes requested):
Looks good, since the assignment block is basically the same ExpressionMap and most transformations simply had transitions as a stop criterion. Please add the backward-compatible json serializer

sdfg/include/sdfg/structured_control_flow/block.h
ramonwirsch: Thats where the Assignment typealias currently lives. We could pull this out into a SDFG / StructuredSDFG agnostic location for Assignments...

opt/src/passes/scheduler/loop_scheduling_pass.cpp
ramonwirsch: Yes. It was a fix for a bug in MapFusionByDomain, where it accidentally renamed indvars in certain circumstances when not fusing, causing 2 loops to have the same indvar (it applied renaming the indvar of the target loop to that of the sourceloop but then not fusing them).
Adrian wrote this pass specifically to make the names unique again. With the MapFusionByDomain now fixed, no current transformation should introduce such reused vars again. And midterm, with SSA its impossible and sth. we need to prevent anyway in all passes.

ramonwirsch:
Backward parser compatibility added. I rechecked the state of transfertuning. Only PolybenchRPCSchedulerTest.Symm fails anymore. But a hell of a lot of polybench_transfertuning_test.py with applying 0 transformations.

ramonwirsch:
Transfertuning was manually tested to work, once we deploy a backend that uses this docc version as basis.

PR #857, opened 2026-07-17
Added PyTorch frontend

Moehre2:
This frontend uses torch.export to lower the GraphModule to Core ATen IR and parses this IR directly. Currently, it can compile resnet18.

Other changes:
- Removed the not-used "CPU version" of the batchnorm tensor node (dead code)
- Updated the Python bindings to support creating more tensor nodes directly
- Added a tensor copy node that can copy one container into another - no matter the tensor types - as long as they have the same number of elements
- Changed the Max, Min, and Sum tensor nodes to expand to reduce nodes instead of for loops
- Also changed the pooling node to use reduce nodes instead of for loops in its expand method


pytorch/tests/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+        for res_elem, ref_elem in zip(res, ref):
+            compare(res_elem, ref_elem, rtol=rtol, atol=atol, equal_nan=equal_nan)
+    else:
+        assert False, f"Unsupported result types: {type(res)} and {type(ref)}"
+
+
+def check(
+    model,
+    *inputs,
+    rtol: float = 1e-4,
```
lukastruemper, line 51: Quite large tolerances. I would prefer to have the same check that we in polybench_*_tests.py for the LLVM frontend. Or at least use testing.assert_allclose(actual, desired, rtol=1e-07, atol=0, equal_nan=True, err_msg='', verbose=True, *, strict=False)

pytorch/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[project]
+name = "docc-pytorch"
+dynamic = ["version"]
+description = "A JIT compiler for PyTorch GraphModules for stateful dataflow multigraphs (SDFG)."
```
lukastruemper: A backend for PyTorch compile based on the Daisytuner Optimizing Compiler Collection (DOCC)

pytorch/docc/pytorch/graph_parser/reshaping.py
```diff
@@ -0,0 +1,84 @@
...
+                "Expected exactly 2 arguments but got " + str(len(node.args)),
+            )
+        if len(node.kwargs) != 0:
+            raise GraphParserError(
+                self, node, "Unsupported kwargs: " + str(node.kwargs)
+            )
+        self.update_container_types(node, builder, container_info, node.name)
+
+
+register_pre_module("aten.permute.default", TensorReshape2dParser())
```
lukastruemper, line 83: Mutating name is somewhat unconvential? Reshape or Views?

pytorch/docc/pytorch/graph_parser/builtin.py
```diff
@@ -0,0 +1,79 @@
...
+    def parse(
+        self,
+        node: torch.fx.Node,
+        builder: StructuredSDFGBuilder,
+        container_info: ContainerInfos,
+    ) -> None:
+        pass
+
+
+register_pre_module("_operator.getitem", GetitemParser())
```
lukastruemper, line 82: Native is not clear to me. It seems like this is the indexing/slicing part? In general, it would be smart to have a class document that clearly states what the module is supposed to parse if the names are not unambiguous.

lukastruemper: Ah, those are the built-in ops like "+, -, *, /". I usually call them Arithmetic, Logical, etc.

One important thing to check is integer division and module for signed and usigned operations (or positive and negative divisors). There are three different ways to do modulo (mod, rem, remainder) and we generate c/c++ code, which has a different behavior than Python, so we need to convert it when translating it. Important to have tests. Cf. https://github.com/daisytuner/docc/blob/b57f184e71540d9777d3537cb046d4d4a3563447/python/tests/python/ast_parser/test_arithmetic.py#L89
https://github.com/daisytuner/docc/blob/b57f184e71540d9777d3537cb046d4d4a3563447/python/tests/python/ast_parser/test_arithmetic.py#L40

lukastruemper (changes requested):
See comments.
Another part that is currently not checked is implicit-casting. NumPy defines rules for implicit-casting, so does PyTorch. There's no test suite here checking this, adapt this test suite for pytorch: https://github.com/daisytuner/docc/blob/main/python/tests/python/functions/numpy/test_numpy_implicit_casts.py

Furthermore, I am not happy with a new pytorch CI-pipeline that defines cuda, rocm steps. Can we not distribute them across the linux workflow with a new pytorch step and then pytorch steps for cuda/rocm as part of the existing steps?

lukastruemper (approved):
Looks great, let's go!

PR #861, opened 2026-07-20
Zero-fill to Memset Pass

Atrisan:


opt/include/sdfg/passes/zero_fill_to_memset_pass.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include "sdfg/analysis/loop_analysis.h"
+#include "sdfg/analysis/memory_layout_analysis.h"
+#include "sdfg/passes/pass.h"
+#include "sdfg/structured_control_flow/map.h"
+#include "sdfg/symbolic/symbolic.h"
+#include "sdfg/types/type.h"
+#include "sdfg/visitor/structured_sdfg_visitor.h"
+
+namespace sdfg::passes {
```
lukastruemper, line 15: Why not nested namespace syntax as 99.9% of all other files?

opt/include/sdfg/passes/zero_fill_to_memset_pass.h
```diff
@@ -0,0 +1,78 @@
...
+ *   upper bound; the nest is perfect (each Map body contains exactly the next level).
+ * - The innermost body is a single Block whose only computation assigns the constant 0
+ *   to one element of the array, indexed exactly by the induction variables of the nest.
+ * - The MemoryLayoutAnalysis tile for the target container at the outermost Map covers the
+ *   whole array contiguously from offset 0 (bounded dimensions match the declared extent,
+ *   and the covered elements form a dense block), guaranteeing the whole array is covered.
+ *   Relying on the layout analysis also transparently handles linearized accesses such as
+ *   `A[i*M + j]`, which are delinearized to the underlying multi-dimensional shape.
+ */
+class ZeroFillToMemsetPass : public sdfg::passes::Pass {
```
lukastruemper, line 36: sdfg::passes:: unnecessary

opt/src/passes/zero_fill_to_memset_pass.cpp
```diff
@@ -0,0 +1,311 @@
...
+            continue;
+        } else {
+            return false;
+        }
+    }
+    return saw_digit;
+}
+
+// Decides whether the tile of accessed elements covers the entire container contiguously
+// starting at offset 0, so that the loop nest is equivalent to a single memset.
```
lukastruemper, line 63: That's not the question: The questions is whether the container is accesses in a contiguous range starting from element zero (which could be relaxed). I.e., you could also just set the first one hundred elements to zero using a loop nest and convert this to a memset, even if the container could hold 200 elements.

A MemoryTile is the convex hull of memory accesses in the body of a loop nest, hence it does not check contiguouity. Thus, the check is wrong.

In essence, you need to check the strides of each individual memlet against the loop bounds and loop strides via memory layout analysis. This check to classify whether a map is elementwise, is a more general useful property that could be added to the for-classification. Maybe be setting a property on maps as @ramonwirsch suggessted, so we have two orthogonal levels of classifcation (parallelism: for, reduce, map) and memory accesses (elementwise and more?)

lukastruemper (changes requested):
See comment: The applicability check for the pass is not sound

PR #862, opened 2026-07-20
prevent MapFusionByDomain fuse transpose-style nested maps

lukastruemper:


ramonwirsch:
Superseded by #873. Testcases copied over.

PR #866, opened 2026-07-22
Add Support for Complex Types

Atrisan:
- adds complex versions of most floating point types
- adds basic tasklets for complex mathmatics


Atrisan:
- **Potential source of the macOS abort in CI:** `complex_computation(...)` infers `operand_prim` from the *first* input edge and assumes it is valid/complex (`sdfg/src/codegen/utils.cpp`, in the block starting with `auto& graph = tasklet.get_parent();`). Please add strict guards before code emission: reject empty in-edge sets, validate every operand is complex for complex tasklets, and enforce consistent operand primitive types (or explicit promotion rules). Right now, malformed or mixed operands can produce invalid generated code and crash at runtime (matching the abort seen at first `@native` arithmetic execution).
- In `complex_computation(...)` (`sdfg/src/codegen/utils.cpp`), can we validate output type compatibility per op kind before emitting statements (complex output for `complex_add/sub/mul/div/neg`, scalar output for `complex_real/imag`, boolean/integer output for `complex_eq/ne` as intended)? This would fail fast with a clear `InvalidSDFGException` instead of generating code that can abort later.
- In `DataFlowDispatcher::dispatch_tasklet` (`sdfg/src/codegen/dispatchers/block_dispatcher.cpp`), please add a defensive invariant near the complex branch (`if (data_flow::is_complex(tasklet.code()))`) that ensures the language extension returns a full statement sequence only for complex tasklets. This protects regular tasklets from accidental non-expression returns and makes regressions easier to catch.
- In `Tasklet::validate` (`sdfg/src/data_flow/tasklet.cpp`), consider adding a corresponding output-type check for complex tasklets (not only input checks). Input validation exists, but output mismatches can still slip through and later fail in codegen/runtime.
- In `primitive_type` classification helpers (`sdfg/include/sdfg/types/type.h`), please make signed/unsigned behavior for complex primitives explicit and consistently documented (tests now expect non-integer behavior). This avoids ambiguous downstream assumptions in codegen and type dispatch paths.

PR #868, opened 2026-07-22
Add tensor concatenation to PyTorch frontend (and MLIR frontend)

Moehre2:
- Added a concat tensor node that concatenates arbitrarily many tensors
- Mapped `torch.ops.aten.cat.default` to concat tensor node
- Changed translation of `tensor.concat` MLIR op to concat tensor node instead of loop nest
- Added replace_symbols method to all types to replace symbols in types like the tensor type
- Added target specific expansion for concat tensor node to avoid branching on GPUs

### DIFF
--- a/opt/include/sdfg/targets/cuda/math/tensor/concat_expander.h
+++ b/opt/include/sdfg/targets/cuda/math/tensor/concat_expander.h
@@ -1,0 +1,30 @@
+#pragma once
+
+#include "sdfg/analysis/analysis.h"
+#include "sdfg/builder/structured_sdfg_builder.h"
+#include "sdfg/data_flow/library_node.h"
+#include "sdfg/data_flow/library_nodes/math/tensor/concat_node.h"
+#include "sdfg/targets/offloading/target_expansion.h"
+
+namespace sdfg {
+namespace offloading {
+
+class CudaConcatExpander : public TargetLibNodeExpander {
+private:
+    math::tensor::ConcatNode& node_;
+
+public:
+    CudaConcatExpander(math::tensor::ConcatNode& library_node)
+        : TargetLibNodeExpander(library_node), node_(library_node) {};
+
+    bool expand(builder::StructuredSDFGBuilder& builder, analysis::AnalysisManager& analysis_manager);
+
+    static bool expand_concat_separately(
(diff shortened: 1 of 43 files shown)


opt/include/sdfg/targets/cuda/math/tensor/concat_expander.h
```diff
@@ -0,0 +1,30 @@
...
+#include "sdfg/analysis/analysis.h"
+#include "sdfg/builder/structured_sdfg_builder.h"
+#include "sdfg/data_flow/library_node.h"
+#include "sdfg/data_flow/library_nodes/math/tensor/concat_node.h"
+#include "sdfg/targets/offloading/target_expansion.h"
+
+namespace sdfg {
+namespace offloading {
+
+class CudaConcatExpander : public TargetLibNodeExpander {
```
ramonwirsch, line 12: Still the old expansion scheme that can only handle 1 expansion per block before a restart is required

sdfg/include/sdfg/data_flow/library_nodes/math/tensor/concat_node.h
```diff
@@ -0,0 +1,89 @@
...
+#include "sdfg/structured_control_flow/block.h"
+#include "sdfg/symbolic/symbolic.h"
+
+namespace sdfg {
+namespace math {
+namespace tensor {
+
+inline data_flow::LibraryNodeCode LibraryNodeType_TensorConcat("ml::Concat");
+
+class ConcatNode : public TensorNode {
```
ramonwirsch, line 29: Code comment on the class.
(does not need to be long and can reference torch docs, but we really should get into the habit of this)

ramonwirsch (approved):

opt/include/sdfg/targets/cuda/math/tensor/concat_expander.h
Moehre2: Is already added to the list of things to do. Didn't have the time...

sdfg/include/sdfg/data_flow/library_nodes/math/tensor/concat_node.h
Moehre2: Ups. Putting that into the next PR.

PR #873, opened 2026-07-23
Adding ProducerIntoConsumer MapFusion into the new, recursive MapFusionPass

ramonwirsch:
Refactored the code of MapFusion Transform into MapFusionByAccessWorker.

The old transform still exists for its use as standalone transformation for now and also to run the old pass to cover everything not yet implemented in the new ByAccessWorker.

Refactored MapFusionByDomain into NewMapFusionPass, the sequential & recursive map-fusion logic into a base that handles the existing by-domain fusion and the by-access fusion, separately disablable.

For this, the existing per-loop caches of NewMapFusionPass have been extended to contain the data needed by by-access fusion (assumptions, non-map loops and more details about subsets and locations of accesses).

NewMapFusionPass still maintains these caches as fusions happen, so it never needs to clear any analysis.

For now, the MapFusionByAccessWorker has not been verified and skips any ConsumerIntoProducer fusions, as that is a greater change to how the cached data needs to be updated and its also a smaller portion of compile-time.

The MapFusionByAccessWorker was refactored to use the FusionLoopCandidate structure to access any cached data. It can be used as a basis for a new Transformation, that simply creates these per-loop structures on the fly, so that Transform and efficient pass can share code again.
The ByAccessWorker was also optimized to support per "fusion container" selection of replacing memory accesses with local transients and to consider the domain-equality checks already done by the NewMapFusionPass.

If the domains match, we can leave in the indirect writes in, which makes the fused code a valid substitute for the original loop, avoiding the need for any loop-duplication.

Elliding same-subset indirect write-read patterns in the same block or successive blocks is a far easier problem to solve, then leaving the original loop lying around and hoping that it can be proven completely dead and removed by later passes.

I.e. fusing loops with indirect accesses has very low downsides and Redundant-Load-Elimination has a very high chance of applying. Duplicating loops has more downsides, so rather a more selective optimization that requires profile-guidance or heuristics to mitigate its downsides.


sdfg/include/sdfg/data_flow/memlet.h
```diff
@@ -351,5 +351,8 @@ class Memlet : public Element {
     void replace(const symbolic::Expression old_expression, const symbolic::Expression new_expression) override;
     void replace(const symbolic::ExpressionMapping& replacements) override;
 };
+
+Subset remap_subset(const Subset& subset, const symbolic::ExpressionMapping& replacements);
```
lukastruemper: Subset is just a typedef for a vector of expressions. Either move it as a util to symbolic or make it part of the class. Free-flying functions are not nice

sdfg/include/sdfg/symbolic/assumptions.h
```diff
@@ -291,5 +295,7 @@ class Assumption {
  */
 typedef std::unordered_map<Symbol, Assumption, SymEngine::RCPBasicHash, SymEngine::RCPBasicKeyEq> Assumptions;
 
+bool replace_indvars_in_assumptions(Assumptions& assumptions, const symbolic::ExpressionMapping& replacements);
```
lukastruemper: static class function? Free-flying function

sdfg/src/symbolic/assumptions.cpp
```diff
@@ -154,7 +154,66 @@ Assumption Assumption::create(const Symbol symbol, const types::IType& type) {
...
+        if (SymEngine::is_a<SymEngine::Symbol>(*replacement)) {
+            auto new_symbol = SymEngine::rcp_static_cast<const SymEngine::Symbol>(replacement);
+            symbol_ = new_symbol;
+            return ReplaceResult::IdChanged;
+        }
+    }
+    return ReplaceResult::IdSame;
+}
+
+bool replace_indvars_in_assumptions(Assumptions& assumptions, const symbolic::ExpressionMapping& replacements) {
```
lukastruemper: What does this have to do specifically with an indvar? There's no specific check. seems like a general helper that's called by someone who decides what an indvar is

python/benchmarks/npbench/weather_stencils/test_hdiff.py
```diff
@@ -66,7 +66,7 @@ def kernel(in_field, out_field, coeff):
 def test_hdiff(target):
     verifier = None
     if target == "none":
-        verifier = SDFGVerification(verification={"SEQUENTIAL": 17, "MAP": 17})
+        verifier = SDFGVerification(verification={"SEQUENTIAL": 20, "MAP": 20})
```
lukastruemper, line 69: Do we know why we lost fusion opportunities?

opt/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ set(SOURCE_FILES
     src/passes/dot_expansion_pass.cpp
     src/passes/einsum.cpp
     src/passes/gemm_expansion_pass.cpp
-    src/passes/map_fusion_by_domain_pass.cpp
+    src/passes/map_fusion/map_fusion_by_access.cpp
+    src/passes/map_fusion/new_map_fusion_pass.cpp
```
lukastruemper: Generally, not a big fan of "new_*" naming because it's not meaningful for anyone not knowing the "old" variant

opt/include/sdfg/passes/normalization/map_fusion.h
```diff
@@ -14,12 +14,15 @@ class MapFusion : public visitor::NonStoppingStructuredSDFGVisitor {
     std::unique_ptr<analysis::LoopAnalysis> loop_analysis_;
     // When false, Case 2 (init-into-reduction hoisting) is disabled for this run.
     bool allow_init_hoist_;
+    // Don't fuse cases of ProducerIntoConsumer
+    bool cons_into_prod_only_;
```
lukastruemper: Minor: Somehow the other way around than expected. "disabled_producer_into_consumer", but con_into_prod_only doesn't tell me that there are only two modes, where the other is disabled

opt/src/passes/loop_fusion/loop_fusion_pass.cpp
```diff
@@ -0,0 +1,1051 @@
...
+                "indvar",
+                {indvar_bounds->tight_lower_bound(), indvar_bounds->tight_upper_bound(), stride, symbolic::integer(level)
+                }
+            );
+        } else {
+            return {};
+        }
+    }
+
+    bool handleStructuredLoop(sdfg::structured_control_flow::StructuredLoop& node) override {
```
lukastruemper, line 110: Name

opt/src/passes/loop_fusion/loop_fusion_by_access.cpp
```diff
@@ -0,0 +1,1114 @@
...
+    }
+    loop_stack.push_back(loop);
+
+    std::ranges::reverse(loop_stack);
+
+    return loop_stack;
+}
+
+std::unique_ptr<MapFusionByAccessWorker::Plan> MapFusionByAccessWorker::
+    try_create_fusion_by_access_plan(FusionLoopCandidate& first, FusionLoopCandidate& second, bool domains_match) {
```
lukastruemper, line 328: Internal note: That's essentially the logic of MapFusion transformation. So we call the "old" Map fusion now "MapFusionByAccess"

lukastruemper: Would have been smart to share the logic between the transformation and the pass by having static functions on the transformation. It's essentially the logic, which is complex equation system solving and needs to be maintained

lukastruemper: Nevermind, this was replaced in the transformation

lukastruemper (changes requested):
First round of review:
- minor parts: naming, free-flying functions, camel case which were hard to read
- MapFusion transformation's symbolic/isl logic has been moved to MapFusionByAccess and then been replaced. So no code duplication which is great.

Please update and I would check fusion by domain afterwards plus integration with @NoraHagmeyer's changes

sdfg/include/sdfg/data_flow/memlet.h
ramonwirsch: I am aware its a typedef. That is why there is no class for it to be a member of, only unrelated classes.

This was closest to be findable. Because if the argument is, you have to know that Subset is just a vector<Expression>, then should we have the typedef at all?

sdfg/include/sdfg/symbolic/assumptions.h
ramonwirsch: Again, there is no "Assumptions" class. Its only a typedef

sdfg/src/symbolic/assumptions.cpp
ramonwirsch: Its only tested for that case and replacing symbols with other symbols, not the full replacement of any expression with another expression.

python/benchmarks/npbench/weather_stencils/test_hdiff.py
ramonwirsch: Yes, we know. Part of the more accurate checks for matching subsets in further nested classes that are not themselve fusing candiates. I believe its sth. like the one loop operates until outer_limit+inner_limit and the consumer loop accesses outer_indvar+1 or sth. so that cannot be supported by by-domain fusing. And the contents are not fit for by-access fusion.

opt/CMakeLists.txt
ramonwirsch: I am aware. What would be your naming suggestion to separate the old from the new one? Its no longer "By-Domain" since it can do both.

opt/include/sdfg/passes/normalization/map_fusion.h
ramonwirsch: "disabled_producer_into_consumer" also does not tell you that there are only 2 modes and the other one is still active either.
But we can invert that flag

opt/src/passes/loop_fusion/loop_fusion_pass.cpp
ramonwirsch: as you can see from that override, its inherited. That naming mistake was made in the middle of last year and tried not to do such breaking changes on the visitor as a by product of my big PR.

sdfg/include/sdfg/data_flow/memlet.h
lukastruemper: add substitute/replace in symbolic.h with std::Vector<Expression>

opt/CMakeLists.txt
lukastruemper: Maybe loop fusion

lukastruemper (approved):

PR #876, opened 2026-07-23
Added mapping for fill to PyTorch frontend

Moehre2:
Also added:
- Restriction to skip tests on older PyTorch versions
- Adapted comparing of results to support numpy


lukastruemper (dismissed):
LGTM

PR #880, opened 2026-07-24
Reduce variability between compile outputs and transfertuning requests

ramonwirsch:
## Description
 * dot_viz sorts edges deterministically. Makes .dot files more deterministic
 + Context can also produce a shared_ptr of the context
 + RPCNodeTransform can optionally dump the cutout and response into sub-folders for debugging (default off, using output_dir metadata on sdfg)
 * updated llvm frontend to use DoccTargets and construct RpcLoopScheduler fresh when needed
 * updated llvm frontend to use the docc context for access to DoccTargets (no it vectorizes again)
 + added "applied_to_element_id" field to RPCNodeTransform reporting, so that we can identify which level of a loop nest was actually tuned
~ replaced global PassReportCollector with thread-safe management that creates reports per module.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Refactor / internal change (no functional change)
- [ ] Documentation update

## Quality Checklist
*Please ensure the following are completed before requesting review:*

* [x] Runtime-compatible (no externally visible existing method has its parameters changed, was deleted or has its exception behavior changed)
* [ ] Compile-time compatible (new arguments & properties added have default values, but users need to recompile against the new version to work)

IR:
* [ ] Introduces new types into the SDFG IR
* [ ]  Changed / added properties on SDFG IR Elements (LibraryNodes, ControlFlowNodes DataFlowNodes etc.)
  *  [ ] New / modified types have serialization / deserialization support
  *  [ ] Deserialization is backward compatible (new serialization can deserialize old format and represent it in the new format if applicable)
  *  [ ] Serialized format is backward compatible (previous deserializer can read new serialized format without incorrectness (just lacking additional guarantees that are optional or did not exist before)
  *  [ ] Serializer has an option to emit the previous format

### Unit Tests
- [ ] New unit tests have been added for the changes.
- [ ] Existing unit tests pass locally.
- [ ] Edge cases and error paths are covered.

### Integration Tests
- [ ] New integration tests have been added (or existing ones updated).
- [ ] Integration tests pass locally.
- [ ] Interactions with other components/services are verified.

### Documentation
- [ ] Public APIs, options, and behavior changes are documented.
- [ ] README / docs site / inline comments updated where applicable.
- [ ] Changelog / release notes updated (if applicable).

### Test Coverage
- [ ] Test coverage has not decreased as a result of this change.
[...]

PR #907, opened 2026-07-29
Send whole sdfg with rpc

NoraHagmeyer:
## Description
Prepares removing the loop normalization and bringing down compile time for segformer with target cuda (with and without remote_tuning) .

- Replaces the RPCScheduler in the LoopSchedulingPass with an RPCSchedulingPass
- Sends the whole sdfg to the RPC server (still using a transformation, to be made more efficient)
- Changes the transer tuning rest endpoint to a newly added one
- Handles response of multiple region informations
- Removes the RPC Opt application

- Changes the result in the Opt Report (to be re-added later or changed to a better mechanism for tracking) regression already tracked through #908 
-  Possible conflicts with #880 

## Related Issues

- Closes #

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Quality Checklist

IR:

### Unit Tests

### Integration Tests

### Documentation

### Test Coverage

## How Has This Been Tested?
RPC unit tests in combination with local rpc server
Transfer Tuning Integration tests in combination with local rpc server
Ran benchmarks for this code state


rpc/src/transformations/rpc_node_transform.cpp
```diff
@@ -40,34 +40,45 @@ std::string RPCNodeTransform::get_node_id_str() const { return std::to_string(th
...
-        if (loopnest_idx >= 0) {
-            report_->in_outermost_loop(loopnest_idx);
+    bool applicable = false;
+    auto outermost_loops = loop_analysis.outermost_loops();
+
+    std::cerr << "Found " << outermost_loops.size() << " outermost loops in node " << get_node_id_str() << std::endl;
+
+    for (auto outermost : outermost_loops) {
+        std::cerr << "[RPC] Checking applicability for node " << get_node_id_str()
+                  << " (loopnest_index=" << loop_analysis.loop_info(outermost).loopnest_index << ")" << std::endl;
```
Moehre2: Should this be `std::cerr`? This means it is also printed in Release mode. If it is only debug output `DEBUG_PRINTLN` should be used.

rpc/src/transformations/rpc_node_transform.cpp
```diff
@@ -260,51 +319,43 @@ void RPCNodeTransform::
...
         auto recipe = opt.local_replay.value();
-        std::cout << "Applied RPC optimization sequence with speedup " << opt.metadata.speedup
-                  << " and vector distance " << opt.metadata.vector_distance << " to loopnest " << element_id << ":\n";
-
-        if (dump_steps_) {
-            print_transformation_sequence(recipe.sequence);
+        for (const auto& region_result : opt.results) {
+            std::cout << "Applied RPC optimization sequence with speedup " << region_result.metadata.speedup
+                      << " and vector distance " << region_result.metadata.vector_distance << " to loopnest "
+                      << element_id << ":\n";
```
Moehre2: Again, should this go to `std::cout` - not `DEBUG_PRINTLN`? (Although this used `std::cout` before)

rpc/src/transformations/rpc_node_transform.cpp
```diff
@@ -260,51 +319,43 @@ void RPCNodeTransform::
...
         }
-
     } else {
-        std::cout << "RPC: Applied plain SDFG with speedup " << opt.metadata.speedup << " and vector distance "
-                  << opt.metadata.vector_distance << "\n";
+        for (const auto& region_result : opt.results) {
+            std::cout << "RPC: Applied plain SDFG with speedup " << region_result.metadata.speedup
+                      << " and vector distance " << region_result.metadata.vector_distance << "\n";
```
Moehre2: See above.

Moehre2 (dismissed):
Not sure if this is intended or slipped through... Therefore, approving.

python/bindings/py_structured_sdfg.cpp
```diff
@@ -495,6 +488,14 @@ void PyStructuredSDFG::schedule(const docc::target::TargetOptions& options) {
         dce.run(builder, analysis_manager);
     }
 
+    if (options.remote_tuning) {
+        std::shared_ptr<sdfg::passes::rpc::RpcContext> context =
+            sdfg::passes::rpc::DaisytunerRpcContext::from_docc_config();
+        sdfg::passes::scheduler::RPCSchedulingPass
+            rpc_scheduling_pass(context, options.target, options.category, nullptr);
```
ramonwirsch: Use the TargetOptions. This more future-proof than unpacking it.

python/bindings/py_structured_sdfg.cpp
ramonwirsch, line 438: code order: if we need this before rpc opt and only do LoopScheduling after, then we can move about "get target specific schedulers code" also to behind the actual rpc opt

ramonwirsch, line 439: btw: is this cleanup  actively harmful for cuda / rocm? it should not be. Then we can simply run that always and need less target-specific code

rpc/include/sdfg/passes/rpc/rpc_scheduling_pass.h
```diff
@@ -0,0 +1,39 @@
...
+#include "sdfg/passes/rpc/rpc_context.h"
+
+namespace sdfg {
+namespace transformations {
+class Recorder;
+}
+namespace passes {
+namespace scheduler {
+
+class RPCSchedulingPass : public Pass {
```
ramonwirsch: Very restrictive name. Since it just hands over the entire SDFG to a configurable endpoint, this is not very "scheduling" specific.

rpc/src/transformations/rpc_node_transform.cpp
```diff
@@ -40,34 +40,45 @@ std::string RPCNodeTransform::get_node_id_str() const { return std::to_string(th
...
+
+    for (auto outermost : outermost_loops) {
+        std::cerr << "[RPC] Checking applicability for node " << get_node_id_str()
+                  << " (loopnest_index=" << loop_analysis.loop_info(outermost).loopnest_index << ")" << std::endl;
+
+
+        auto loop_info = loop_analysis.loop_info(outermost);
+
+        if (!loop_info.has_side_effects) {
+            std::cerr << "Loop has no side effects, marking as applicable" << std::endl;
```
ramonwirsch: here as well

rpc/src/transformations/rpc_node_transform.cpp
ramonwirsch, line 29: What does this node represent, now that its whole-sdfg? Is it just sdfg.root()?
Then why have it at all?

rpc/include/sdfg/passes/rpc/rpc_scheduling_pass.h
```diff
@@ -0,0 +1,39 @@
...
+private:
+    std::shared_ptr<rpc::RpcContext> rpc_context_;
+    std::string target_, category_;
```
ramonwirsch: More future proof: use TargetOptions instead and for as long as possible.

rpc/src/transformations/rpc_node_transform.cpp
NoraHagmeyer: Yes, it is just root.

python/bindings/py_structured_sdfg.cpp
NoraHagmeyer: Openend an issue for evaluation. Too many moving parts, I want to make as few changes as possible to keep problems contained

rpc/src/transformations/rpc_node_transform.cpp
NoraHagmeyer: We don't need a transform. I opened an issue to go from a transform to just doing it in the pass. The transform is really legacy.

ramonwirsch (dismissed):

ramonwirsch (dismissed):

ramonwirsch (approved):

PR #910, opened 2026-07-29
[PyTorch] Add Support for Embedding Layers

Atrisan:
## Description
Add support for embeddings includes support for the following aten ops:
* embedding.default
* embedding_renorm.default
* alias.default
* detach.default

Adds the following new tensor nodes:
* embedding
* embedding renorm

Only return user outputs, to avoid conflicts with inplace ops like embedding_renorm.default.

## Related Issues

- Closes #

## Type of Change
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Quality Checklist

IR:
* [x]  Changed / added properties on SDFG IR Elements (LibraryNodes, ControlFlowNodes DataFlowNodes etc.)
  *  [x] New / modified types have serialization / deserialization support
  *  [x] Deserialization is backward compatible (new serialization can deserialize old format and represent it in the new format if applicable)
  *  [x] Serialized format is backward compatible (previous deserializer can read new serialized format without incorrectness (just lacking additional guarantees that are optional or did not exist before)
  *  [x] Serializer has an option to emit the previous format

### Unit Tests

### Integration Tests

### Documentation
- [x] README / docs site / inline comments updated where applicable.

### Test Coverage

## How Has This Been Tested?
Unit tests for libnodes.
Integration tests for ops.

## Additional Context


sdfg/src/data_flow/library_nodes/math/tensor/embedding_node.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    if (weight_shape_.size() != 2) {
+        throw InvalidSDFGException(
+            "EmbeddingNode: weight must be 2-dimensional but got rank " + std::to_string(weight_shape_.size())
+        );
+    }
+    if (index_shape_.empty()) {
+        throw InvalidSDFGException("EmbeddingNode: index_shape must not be empty");
+    }
+}
```
WashimNeupane, line 42: Maybe add connector type check here since we are skipping TensorNode::validate(). At minimum, it should verify that "I" edge is an integer type and "Y/W" edge are float/non-integer

pytorch/docc/pytorch/graph_parser/__init__.py
```diff
@@ -66,12 +66,21 @@ def get_output_containers(self, node: torch.fx.Node, args: Argument) -> list[str
...
         """
         result = []
         if isinstance(args, tuple):
             for elem in list(args):
                 result += self.get_output_containers(node, elem)
         else:
+            if isinstance(args, torch.fx.Node):
+                user_outputs = getattr(self.ep.graph_signature, "user_outputs", None)
+                if user_outputs is not None and args.name not in user_outputs:
+                    return result
```
WashimNeupane, line 84: The check is written generically for any output node that isn't in user_outputs. If someone adds another op tomorrow that unintentionally lands outside user_outputs (perhaps due to a bug), this code will quietly discard it instead of raising an error, making the bug very hard to find. maybe we have to raise an error when an unknown non-user-visible output is encountered.

sdfg/src/data_flow/library_nodes/math/tensor/embedding_node.cpp
Atrisan: I is reasonable, but this tensor node support int as well so Y/W can be int and float.
See line 25:
bool EmbeddingNode::supports_integer_types() const { return true; }

pytorch/docc/pytorch/graph_parser/__init__.py
Atrisan: This is exactly the idea.
Because embedding_renorm is an inplace operation you run for its side effect.
As a result, its input will be considered an output as well, for the entire model.
Even though it is not a real output.

The problem is that you need to filter those to ensure analogous outputs to plain PyTorch.
I'll have to check if this becomes redundant, when the output reconstruction is merged.
Until then, we need to ensure that we do not add new tupel elements to the output.

I am also not sure if this is the best solution to this problem either.

Moehre2 (approved):

PR #917, opened 2026-07-30
Add scalar tensor

WashimNeupane:
## Description
Added a aten.scalar_tensor op to frontend. 
The requirement to have a persistent session across the SDFG was removed/commented out. Need better fix in another PR. 

## Related Issues

- Closes #

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Quality Checklist

IR:

### Unit Tests

### Integration Tests

### Documentation

### Test Coverage

## How Has This Been Tested?
-verified no memory allocation (malloc) is removed for the 0d stack-allocated scalar tensor.

## Additional Context

### DIFF
--- a/pytorch/docc/pytorch/graph_parser/creation.py
+++ b/pytorch/docc/pytorch/graph_parser/creation.py
@@ -153,7 +153,7 @@ class ScalarTensorParser(GraphParserModule):
                 f"Expected exactly 1 argument (scalar value), got {len(node.args)}",
             )
         if not set(node.kwargs.keys()).issubset(
-            {"dtype", "layout", "device", "pin_memory", "memory_format"}
+            {"dtype", "layout", "device", "pin_memory"}
         ):
             raise GraphParserError(self, node, f"Unsupported kwargs: {node.kwargs}")
 
@@ -242,20 +242,11 @@ class ScalarTensorParser(GraphParserModule):
 
         container_type = container_info[result_container].sdfg_type()
         debug_info: DebugInfo = self.get_debug_info(node)
-        if isinstance(container_type, Pointer):
-            builder.add_fill_op(
-                fill_value_container,
-                fill_value_type,
-                result_container,
-                result_tensor,
-                debug_info,
-            )
-        else:
-            builder.add_assignment(
-                result_container,
-                fill_value_container,
-                debug_info,
-            )
+        builder.add_assignment(
+            result_container,
+            fill_value_container,
+            debug_info,
+        )
 
 
 register_module("aten.scalar_tensor.default", ScalarTensorParser())
--- a/pytorch/docc/pytorch/graph_parser/creation.py
+++ b/pytorch/docc/pytorch/graph_parser/creation.py
@@ -5,7 +5,7 @@ GraphParser modules for parsing operations to create tensors.
 import torch.fx
 from torch.fx.node import Argument
 
-from docc.sdfg import StructuredSDFGBuilder, Scalar, Tensor, DebugInfo
+from docc.sdfg import StructuredSDFGBuilder, Scalar, Pointer, Tensor, DebugInfo
 
 from docc.pytorch.graph_parser.utils import (
     GraphParserModule,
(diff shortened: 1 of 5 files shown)


pytorch/tests/ops/test_creation.py
```diff
@@ -49,3 +49,46 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
...
+
+    check(ScalarTensorLayoutNet(), target=target)
+
+
+def test_scalar_tensor_add(target: str) -> None:
+    class ScalarTensorAddNet(nn.Module):
+        def forward(self, input: torch.Tensor) -> torch.Tensor:
+            return input + torch.scalar_tensor(3.141592)
+
+    check(ScalarTensorAddNet(), torch.ones(2, 3), target=target)
```
Atrisan, line 94: What about tests for pinned_memory and memory_format?
Are they not supported?

pytorch/docc/pytorch/graph_parser/utils.py
```diff
@@ -899,13 +899,20 @@ def allocate_memory(
                 node,
                 "Expected ContainerInfo but got: " + str(type(info)),
             )
+
+        # Skip allocation for Scalar Tensors
+        if isinstance(info.sdfg_type(), Scalar):
```
Atrisan: You are checking for scalars, which should not be allocated.
A Scalar tensor is still of type Tensor, which is why it needs additional coverage.

If a scalar is passed to this function, something is not filtered correctly in the parser.
Thus, if you add this check it should rather throw an error than hide a noop.

pytorch/docc/pytorch/graph_parser/creation.py
```diff
@@ -117,15 +117,91 @@ def parse(
                 node, fill_value, result_tensor.element_type
             )
 
+        container_type = container_info[result_container].sdfg_type()
+        if isinstance(container_type, Pointer) and len(result_tensor.shape) == 0:
```
Atrisan: Why do you check for pointers to create a new tensor?
Either this is something to handle in the utils or its somewhat redundant.
It looks fishy

pytorch/docc/pytorch/graph_parser/creation.py
```diff
@@ -117,15 +117,91 @@ def parse(
...
+        container_info: ContainerInfos,
+    ) -> None:
+        if len(node.args) != 1:
+            raise GraphParserError(
+                self,
+                node,
+                f"Expected exactly 1 argument (scalar value), got {len(node.args)}",
+            )
+        if not set(node.kwargs.keys()).issubset(
+            {"dtype", "layout", "device", "pin_memory", "memory_format"}
```
Atrisan: kwarg tests are never type checked like the FullParser

Atrisan (changes requested):

pytorch/tests/ops/test_creation.py
WashimNeupane: i was following other tests patterns i.e. fullparser.. they are supported.

pytorch/docc/pytorch/graph_parser/utils.py
WashimNeupane: When we use a 0-dimensional tensor internally in PyTorch module (like input + torch.scalar_tensor(...) in test_scalar_tensor_add), the docc graph parser realizes it doesn't need to return it, so it maps it to a primitive Scalar container. This guard was to avoid that. However, new changes to the code will not map it to scalar container within this function, so this wont be necessary.

pytorch/docc/pytorch/graph_parser/creation.py
WashimNeupane: we were initialising a pointer for scalar tensor. this causes weird bugs. this will be resolved by simply always creating a Scalar container (no pointers), which gets changed to a pointer bufffer when required.

pytorch/docc/pytorch/graph_parser/creation.py
WashimNeupane: added to new commit.

Atrisan (dismissed):

Atrisan (approved):

PR #918, opened 2026-07-30
Loop fusion migration in normalize

ramonwirsch:
## Description
Adds the quicker LoopFusion to the normalize stage. 
As in simplify, ConsumerIntoProducer cases / backward fusing is not supported yet, so the old MapFusion pipeline is still run after it. But since the LoopFusion can fuse most of it, it saves a lot of compile time on larger projects.

LoopFusion was slightly improved to avoid performance regressions:
 * Loop fusion of 2 loops with no shared data structures is restricted to cases where all loop levels match. This prevents us from fusing outer levels alone, which prevented other optimizations down the road.

MapFusion is restricted from handling ProducerIntoConsumer cases, same as in simplify, as 99% of them should be handled by the LoopFusion.

## Type of Change
- [*] Refactor / internal change (no functional change)
- [ ] Documentation update

## Quality Checklist

* [*] Runtime-compatible (no externally visible existing method has its parameters changed, was deleted or has its exception behavior changed)
* [*] Compile-time compatible (new arguments & properties added have default values, but users need to recompile against the new version to work)

IR:

### Unit Tests
- [*] New unit tests have been added for the changes.
- [*] Existing unit tests pass locally.
- [*] Edge cases and error paths are covered.

### Integration Tests

### Documentation

### Test Coverage

### DIFF
--- a/llvm/integration/polybench_cuda_test.py
+++ b/llvm/integration/polybench_cuda_test.py
@@ -402,12 +402,12 @@ def test_symm(datatype, compiler="clang-19", size="MEDIUM_DATASET"):
     verifier = SDFGVerification(
         verification={
             "sdfgs": 8,
-            "REDUCE": 4,
+            "REDUCE": 3,
             "CUDAOffloading": 8,
             "MAP": 7,
             "CUDA": 4,
             "SEQUENTIAL": 10,
-            "FOR": 3,
+            "FOR": 4,
         },
     )
     test_case = benchmark_path / "symm.c"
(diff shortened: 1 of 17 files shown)


NoraHagmeyer (dismissed):

NoraHagmeyer:
Verifiers need another update but looks good to me otherwise!

NoraHagmeyer (changes requested):
Polybench verifier changes are looking good. Can you also run the apps locally and adapt the verifiers accordingly? Otherwise main will be failing after your merge.

NoraHagmeyer (approved):

PR #924, opened 2026-07-31
[PyTorch] Reconstruct output args to fit pytorch and other bug fixes

Atrisan:
## Description
Reorders result tuples to fit the original pytorch output.
Generate copy kernels where the strides and tensor layout are regenerated instead of being ignored.
Fix the 2 broadcast index to the correct dimension.
Skip dynamic size parsing with torch compile.

## Related Issues

- Closes #

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Quality Checklist

IR:

### Unit Tests

### Integration Tests

### Documentation
- [x] README / docs site / inline comments updated where applicable.

### Test Coverage

## How Has This Been Tested?

## Additional Context


pytorch/docc/pytorch/graph_parser/convolution.py
```diff
@@ -60,14 +60,31 @@ def parse(
...
         result_tensor = self.get_tensor_type(node, container_info, result_container)
         debug_info: DebugInfo = self.get_debug_info(node)
+        # DOCC's convolution expansion always writes a contiguous (NCHW) output.
+        # When torch.compile selects a non-contiguous memory format for the
+        # convolution (e.g. channels_last), the fx metadata expresses the result
+        # -- and every downstream view/permute derived from it -- relative to that
+        # layout. Emit the convolution into a contiguous buffer and copy it into
+        # the requested (strided) layout so the physical data matches what the
+        # consumers assume; otherwise the layout mismatch silently transposes the
+        # data (observed as a wrong LayerNorm result in SegFormer).
```
WashimNeupane, line 70: creating a intermediate tenosr will cause performance/memory issues down the line. the fix is ok for now for correctness test, but a better alternative would be to change the conv library node to accept non-contiguous output i.e. move away from expanding to blas and into cuDNN, MIOpen, ACL, etc.

Atrisan: Not necessarily, thanks to the map fusion many of these interim results and manifestations can actually be fused into their predecessor or successor, as long as they at least one of them expands into a loopnest instead of another libnode. A different lib does not really help here.

sdfg/include/sdfg/data_flow/library_nodes/math/tensor/broadcast_node.h
```diff
@@ -14,6 +14,8 @@ class BroadcastNode : public TensorNode {
 private:
     std::vector<symbolic::Expression> input_shape_;
     std::vector<symbolic::Expression> output_shape_;
+    // true: leading-alignment (linalg.broadcast); false: trailing-alignment (NumPy/PyTorch).
+    bool padded_;
```
WashimNeupane, line 18: maybe something like is_left_aligned would be more appropriate/intuitive than padded_ here.

pytorch/docc/pytorch/graph_parser/convolution.py
WashimNeupane: makes sense.

WashimNeupane (dismissed):

.github/workflows/unit_tests_linux.yml
```diff
@@ -5,11 +5,29 @@ on:
...
-    types: [opened, reopened, synchronize, ready_for_review]
+    types: [opened, reopened, synchronize]
   schedule:
     - cron: "0 4 * * *"
 
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      python-versions: ${{ steps.matrix.outputs.python-versions }}
```
lukastruemper: I feel like this is a bit overcomplicating things. Just split the workloads. Maybe as a separate PR?

lukastruemper (changes requested):

.github/workflows/unit_tests_linux.yml
Atrisan: A separate PR would be reasonable.
The idea behind this was to keep the tests identical and a oid replication. Thereby, encuraging that the tests cannot diverge in future changes as well.

pytorch/tests/__init__.py
```diff
@@ -72,6 +72,14 @@ def check(
...
         model,
         backend="docc",
+        # Specialize on concrete input shapes. Without this, Dynamo's automatic
+        # dynamic-shape behavior turns the batch dimension into a symbolic SymInt
+        # on the second compile of the same model code (e.g. the batch_size=1 and
+        # batch_size=4 parametrizations of the same test). That makes the traced
+        # example_input shapes identical (s0, ...) across batch sizes, so the
+        # compile cache key / stable_id collide and both batches reuse the same
+        # SDFG and .so. Forcing static shapes gives each batch size its own graph.
+        dynamic=False,
```
lukastruemper, line 82: I thought this is what we want, dynamic sizes?

lukastruemper (approved):
LGTM, but thought we want dynamic sizes

pytorch/tests/__init__.py
Atrisan: I agree, but we currently do not support them and if we calculate the same hash, pytorch automatically defaults to dynamic sizes, causing issues in testing.
Once we have support it can be reactivated, but this keeps the behaviour consistent.

PR #938, opened 2026-08-05
Better compile stats that supported nested timings

ramonwirsch:
Replaced PassStatistics, PipelineStatistics, AnalysisStatistics with CompileStatistics that simply records nested start and stop times. The default report summarized deeply nested data to about the same format as before, but DOCC_STATISTICS=2 enables the full report, showing nested calls of analyses the exact flow of each pipeline etc.

Also supports KV metrics that can be added to each entered scope and will be shown as well (like fuse-count)

Also throws in support for tests to find the output folder to dump other data.

### DIFF
--- a/opt/src/passes/normalization/normalize.cpp
+++ b/opt/src/passes/normalization/normalize.cpp
@@ -15,7 +15,7 @@ namespace passes {
 namespace normalization {
 
 void normalize(sdfg::StructuredSDFG& sdfg, bool enable_fusion) {
-    sdfg::passes::CompileStatistics::enter_stage_if_enabled("normalize");
+    CompileStatistics::enter_stage_if_enabled("normalize");
     builder::StructuredSDFGBuilder builder(sdfg);
     analysis::AnalysisManager analysis_manager(sdfg);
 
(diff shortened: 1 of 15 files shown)


opt/src/passes/normalization/normalize.cpp
```diff
@@ -15,25 +15,25 @@ namespace passes {
 namespace normalization {
 
 void normalize(sdfg::StructuredSDFG& sdfg, bool enable_fusion) {
+    sdfg::passes::CompileStatistics::enter_stage_if_enabled("normalize");
```
lukastruemper: Why have namespace qualifiers? sdfg::passes::

opt/tests/transformations/offloading/cuda_parallelize_nested_map_test.cpp
```diff
@@ -739,6 +740,116 @@ TEST(CUDANestedParallelismTransformation, SiblingAccumulationBlocksNestedParalle
     EXPECT_FALSE(transformation.can_be_applied(builder, analysis_manager));
 }
 
+/**
+ * Reproduced a problem seen in npbench gesummv, where a map contains a nested map-reduce constellation that does not
+ *get fused into a single reduce-loop
+ **/
+TEST(CUDANestedParallelismTransformation, SiblingMismatchedIterationsRejected) {
```
lukastruemper: How did this test end up here?

lukastruemper (changes requested):

opt/tests/transformations/offloading/cuda_parallelize_nested_map_test.cpp
ramonwirsch: I still had it locally from last week and decided better to commit it instead of deleting it.
By the time I had written the test, a fix was already in main, it was not worth it to go through the merge process. But its tests for a different issue then the test we have

As mentioned in my PR description.

opt/src/passes/normalization/normalize.cpp
ramonwirsch: I am sure I copied it over from some place that needed it and did not follow my IDE highlights that it was redundant, because that is a common occurrance. Also, the entire file looks that way, so hardly noticeable

lukastruemper (approved):

PR #946, opened 2026-08-10
adds instrumentation reset in harness

lukastruemper:
## Description
A clear and concise description of what this PR does and why.

## Related Issues

- Closes #

## Type of Change
- [ ] Documentation update

## Quality Checklist

IR:

### Unit Tests
- [ ] New unit tests have been added for the changes.
- [ ] Existing unit tests pass locally.
- [ ] Edge cases and error paths are covered.

### Integration Tests

### Documentation

### Test Coverage

## How Has This Been Tested?

## Additional Context

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,16 +73,6 @@ endif()
 add_subdirectory(sdfg)
 add_subdirectory(opt)
 add_subdirectory(rpc)
-
-# The python/mlir frontends need a single shared RTL instance per process so the
-# runtime bindings can reset/query it (see rtl/CMakeLists.txt). Decide the RTL
-# linkage before the rtl subdirectory is configured below.
-option(PYTHON_BUILD_FRONTEND "Build Python frontend" OFF)
-option(MLIR_BUILD_FRONTEND "Build MLIR frontend" OFF)
-if(PYTHON_BUILD_FRONTEND OR MLIR_BUILD_FRONTEND)
-    set(DAISY_RTL_SHARED ON CACHE BOOL "Build the daisy RTL as a shared library" FORCE)
-endif()
-
 add_subdirectory(rtl)
 add_subdirectory(c-compile)
 add_subdirectory(arg-capture-io)
@@ -94,10 +84,12 @@ if(LLVM_BUILD_FRONTEND)
     add_subdirectory(llvm)
 endif()
 
+option(PYTHON_BUILD_FRONTEND "Build Python frontend" OFF)
 if(PYTHON_BUILD_FRONTEND)
     add_subdirectory(python)
 endif()
 
+option(MLIR_BUILD_FRONTEND "Build MLIR frontend" OFF)
 if(MLIR_BUILD_FRONTEND)
     add_subdirectory(mlir)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,16 @@ endif()
 add_subdirectory(sdfg)
 add_subdirectory(opt)
 add_subdirectory(rpc)
+
+# The python/mlir frontends need a single shared RTL instance per process so the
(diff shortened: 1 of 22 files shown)


rtl/CMakeLists.txt
```diff
@@ -4,7 +4,17 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
-add_library(daisy_rtl STATIC
+# Static by default (each artifact embeds its own instrumentation state). The
```
ramonwirsch: This makes llvm frontend's use of rtl / instrumentation no longer work if DAISY_RTL_SHARED is on.

Meaning, we now require 2 separate build folders, one for python-... and one for llvm builds.

I suggest we either always build both and have each front end do what it wants, or we upgrade the llvm frontend to using the dynamic library always as well.

CMakeLists.txt
```diff
@@ -73,6 +73,16 @@ endif()
 add_subdirectory(sdfg)
 add_subdirectory(opt)
 add_subdirectory(rpc)
+
+# The python/mlir frontends need a single shared RTL instance per process so the
+# runtime bindings can reset/query it (see rtl/CMakeLists.txt). Decide the RTL
+# linkage before the rtl subdirectory is configured below.
+option(PYTHON_BUILD_FRONTEND "Build Python frontend" OFF)
```
ramonwirsch: this will implicitly make llvm frontend fail partially by enabling other frontends

c-compile/src/compile/src_file_compiler.cpp
```diff
@@ -278,6 +278,9 @@ bool SrcFileCompiler::run_compile_and_link_single(const std::filesystem::path& s
 void SrcFileCompiler::add_link_args(std::stringstream& cmd) const {
     for (auto& ld : this->library_paths_) {
         cmd << "-L " << ld << " ";
+        // Embed an rpath so a shared RTL (DAISY_RTL_SHARED) resolves at runtime
```
ramonwirsch: This is somewhat dirty as we would also add target-specific libraries to the rpath, that we so far expect to have on LD_LIBRARY_PATH (and that users may expect to switch around after compiling)

Ok to merge, but a cleaner solution would be to have an additional flag that is set for daisy_rtl if we want to link it dynamically or for our own libs.

python/docc/benchmarks/rtl.py
```diff
@@ -54,6 +61,140 @@
...
+    """(exact filename, glob) for the RTL on the current platform."""
+    if sys.platform == "darwin":
+        return "libdaisy_rtl.dylib", "libdaisy_rtl*.dylib"
+    if sys.platform.startswith("win"):
+        return "daisy_rtl.dll", "daisy_rtl*.dll"
+    # Match the plain name and, defensively, any soname-versioned file.
+    return "libdaisy_rtl.so", "libdaisy_rtl.so*"
+
+
+def _candidate_lib_dirs() -> List[Path]:
```
ramonwirsch, line 81: C++ code discovers and tracks the various locations for its binaries and sources and uses those paths to compile and link the SO. So we do not need to try all paths from python, we could just ask the docc native code, which folders it is using.

Ok to merge, but there is a cleaner solution.

ramonwirsch (changes requested):
Makes it impossible to use llvm and python- frontends from the same cmake build setup.
(at least when instrumentation is used).

We can either make the llvm frontend also support the daisy_rtl.so (by doing the same rpath) and make it default & permanently shared. Or we can setup cmake to build both and drop the option to switch which one we build. Then llvm can default to the old way and python frontends can default to the new way.

ramonwirsch (approved):
Resolved all concerns. Using unified shared library + rpath everywhere, with explicit management of rpath and using the paths of library in the source repository we also use for linking if we detect being run from the cmake source tree

PR #956, opened 2026-08-13
fixes bugs in post-RPC scheduling and cleanup

lukastruemper:
## Description
- RPC scheduling does not report loop nests changed, so post-cleanup is never triggered
- Transforms/schedulers do not respect already-scheduled loops from RPC, so the get double offloaded

## Related Issues

- Closes #

## Type of Change
- [ ] Documentation update

## Quality Checklist

IR:

### Unit Tests
- [ ] New unit tests have been added for the changes.
- [ ] Existing unit tests pass locally.
- [ ] Edge cases and error paths are covered.

### Integration Tests

### Documentation

### Test Coverage

## How Has This Been Tested?

## Additional Context

### DIFF
--- a/opt/src/passes/scheduler/cuda_scheduler.cpp
+++ b/opt/src/passes/scheduler/cuda_scheduler.cpp
@@ -23,6 +23,10 @@ SchedulerAction CUDAScheduler::find(
     structured_control_flow::StructuredLoop& loop,
     bool offload_unknown_sizes
 ) {
+    if (loop.schedule_type().category() != structured_control_flow::ScheduleTypeCategory::None) {
+        return SKIP;
+    }
+
     if (dyn_cast<structured_control_flow::Map*>(&loop)) {
         return NEXT;
     }
--- a/opt/src/passes/scheduler/cuda_scheduler.cpp
+++ b/opt/src/passes/scheduler/cuda_scheduler.cpp
@@ -23,12 +23,8 @@ SchedulerAction CUDAScheduler::find(
     structured_control_flow::StructuredLoop& loop,
     bool offload_unknown_sizes
 ) {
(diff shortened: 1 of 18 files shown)


NoraHagmeyer:
I fixed bugs related to double offloading a month ago and it worked for the tests I had. There are also unit tests ensuring that.

opt/src/passes/scheduler/cuda_scheduler.cpp
```diff
@@ -23,6 +23,10 @@ SchedulerAction CUDAScheduler::find(
     structured_control_flow::StructuredLoop& loop,
     bool offload_unknown_sizes
 ) {
+    if (loop.schedule_type().category() != structured_control_flow::ScheduleTypeCategory::None) {
```
NoraHagmeyer: Add this check to loop_schedulining pass and return found_incompatible = true if a loop already has a schedule type. This way, it will work for all current and future schedulers.

sdfg/include/sdfg/passes/scheduler/loop_scheduler.h
```diff
@@ -16,6 +16,7 @@ namespace scheduler {
 enum SchedulerAction {
     NEXT,
     CHILDREN,
+    SKIP,
```
NoraHagmeyer: So far, NEXT could also mean SKIP if nothing was done and then gone to the NEXT loop nest. If you introduce a difference, this logic should also be changed for consistency in all schedulers.

NoraHagmeyer (changes requested):
Thanks for fixing that! I though I already did but seem to have only done so for multi-dimensional offloading.

Ibelieve the loop_scheduling_pass is the better location for the check though.

PR #962, opened 2026-08-18
Migrate to llvm21

ramonwirsch:
We want clang-21 for its support of rocm7 and cuda13, But to avoid conflicts and also update the llvm frontend, we need to migrate that to llvm-21, so that we can use one unified installation to build all our artifacts.

At least on ubuntu, libomp is otherwise conflicting and cannot be installed in both the 19 and 21 versions.

Migration in the front-end was simple thus far, only very slight API changes.
Biggest changes so far have been that the print functions are no longer inlined for all the polybench tests, everything else has stayed the same.

! Currently etsoc tests are deactivated, because the etsoc images do not built (upstream is incomplete).

### DIFF
--- a/.daisy/mlir_torch_models.yml
+++ b/.daisy/mlir_torch_models.yml
@@ -6,7 +6,7 @@ on:
     types: [opened, reopened, synchronize]
 
 parameters:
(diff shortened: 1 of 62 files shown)


lukastruemper (dismissed):
Did a pass, LGTM

PR #989, opened 2026-08-25
PyTorch tensor/container refactor

Moehre2:
- Complete refactor of the tensor/container metadata in the PyTorch frontend
- Moved pip packages from workflows into requirements.txt files
- More code reuse through helper functions for common kwargs
- "Fixed" slicing by removing the SliceNode and performing the op "virtually" instead, i.e., using shape, strides, and offset to access a subset of the data from the same container
- Complete rewrite of the LayerNorm node to make it more aligned with PyTorch and enable the support of non-contiguous input/output tensors
- Moved im2col convolution expansion into a target specific expand for target sequential and openmp. The default expansion of a convolution is now the naïve one. Also updated all convolution expansion to the new expansion format.
- Added a segformer test on an actual image
- Moved tests with more than op into their own folder
- Fixed a bug in the MLIR frontend where buffers were not correctly (re-)used because linalg_custom ops did not implement the DestinationStyleOpInterface
- Disabled two local buffer reuse passes because they relied on the im2col expansion of the convolution and are not correct with the naïve expansion anymore


mlir/lib/Target/SDFG/SDFGTranslator.cpp
```diff
@@ -471,7 +474,7 @@ std::string SDFGTranslator::
             this->builder_.add_computational_memlet(block, out_access, fill_node, "Y", {}, *sdfg_tensor, deb_info);
         } else {
             auto& src_type = builder_.subject().type(output_container);
-            auto& dst_type = builder_.subject().type(copy_container);
+            // auto& dst_type = builder_.subject().type(copy_container);
```
Atrisan, line 477: delete?

opt/src/targets/omp/math/tensor/conv_expander.cpp
```diff
@@ -0,0 +1,391 @@
...
+namespace sdfg {
+namespace omp {
+
+passes::LibNodeExpander::ExpandOutcome OMPConvExpander::handle_expand(
+    passes::LibNodeExpander::ExpandContext& context, structured_control_flow::Block& block, math::tensor::ConvNode& node
+) const {
+    return handle_expand_im2col(context, block, node);
+}
+
+passes::LibNodeExpander::ExpandOutcome OMPConvExpander::handle_expand_im2col(
```
Atrisan, line 26: Nitpick: Can we have those implementation as part of the conv node instead of a specific expander.
And then just have the logic on which one to use in the expander.

Unless there is a very special setup for a given target.

opt/src/targets/rocm/math/tensor/conv_expander.cpp
```diff
@@ -1,21 +1,18 @@
...
 namespace sdfg {
 namespace offloading {
 
-bool RocmConvExpander::expand(builder::StructuredSDFGBuilder& builder, analysis::AnalysisManager& analysis_manager) {
+passes::LibNodeExpander::ExpandOutcome RocmConvExpander::handle_expand(
+    passes::LibNodeExpander::ExpandContext& context, structured_control_flow::Block& block, math::tensor::ConvNode& node
+) const {
     // for now we reuse the cuda impl until we find sth. where they need to diverge
-    return CudaConvExpander::expand_conv(builder, analysis_manager, node_);
+    return CudaConvExpander::handle_expand_im2row(context, block, node);
```
Atrisan: If rocm and cuda, behave identically.
Can we instead use a GPU expander, instead of defining a dependency between cuda and rocm.

Atrisan (changes requested):

mlir/lib/Target/SDFG/SDFGTranslator.cpp
Moehre2: Depends on another comment...

opt/src/targets/omp/math/tensor/conv_expander.cpp
Moehre2: We can discuss this with more people and decide on a general strategy. Then, we can do this for ALL expanders.

opt/src/targets/rocm/math/tensor/conv_expander.cpp
Moehre2: Done.

Atrisan (approved):

PR #990, opened 2026-08-25
Async Kernel Runtime capture for ROCm & CUDA in aggregate mode

ramonwirsch:
Daisy-rtl will switch to using CUDA events on cuda kernels, because measuring runtime on host will not work well.
Same for ROCm
Disabled in non-aggregate mode, due to us needing also a starttime and not getting a unified start time ouf of the event APIs.

__DAISY_INSTRUMENTATION_ASYNC_CAPTURE=0 overrides default and disables this also for aggregate mode.

Instroduced __daisy_instrumentation_finalize_all at the end of an SDFG, to give rtl lib time to wait on and capture any outstanding async events not captured already (will block here until all captured kernels complete...)


lukastruemper (approved):
LGTM

PR #992, opened 2026-08-26
Rocm profiling

ramonwirsch:
Initialize daisy_rtl explicitly from our python code, so it can prepare rocm instrumentation before anyone else initalizes rocm without instrumentation support.

daisy_rtl will now also verify the existence of all counters it is supposed to track during initialization.

This should make most of our benchmark cases work with rocm profiling

### DIFF
--- a/mlir/benchmarks/harness.py
+++ b/mlir/benchmarks/harness.py
@@ -1,13 +1,13 @@
 import argparse
 import os
 import sys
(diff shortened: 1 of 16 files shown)


lukastruemper (approved):
LGTM

Moehre2 (approved):

PR #996, opened 2026-08-26
Extracted bad recipe into unit test

Moehre2:
Old OutLocalStorage procuded wrong code on this. New LocalStorage does not. Added unit test.


lukastruemper (approved):
LGTM

PR #1017, opened 2026-09-01
Reduced indvar sizes

ramonwirsch:
More broadly replace hardcoded Int64 and UInt64 indvars with a function inferring the type from the expression (although for now, it only uses Int32 if it can prove that the immediate result fits into it, otherwise Int64).

Noticed that many places use UInt64, which is not technically valid and potentially problematic. Now enforcing no unsigned indvars on StructuredLoop validator and fixing test cases


lukastruemper (approved):
LGTM

PR #1018, opened 2026-09-01
adds option registry for passes and analyses

lukastruemper:


Moehre2 (dismissed):

ramonwirsch (commented):
Lots of workarounds that require ugly code and have ugly life-cycle implications to get at the options in a way the API of existing passes does not have to change to access them.

Now, every pass needs to sift through the kv options on every run and needs to override options that already exist in the constructor with temporary options during a specific run.

The Options stuct themselves are fine. I just would handle them through the Constructor of passes and require a pass be constructed for a specific set of options.

ramonwirsch (approved):

PR #1023, opened 2026-09-02
Fix LLVM test suite

Moehre2:
Fix LLVM test suite test for LLVM frontend by moving the test cases to our benchmark-tests repo and removing the logic to checkout the git repository from the pytest.


NoraHagmeyer (approved):
We may want to reevaluate again at some point since checking out the repository itself is nicer. But it's a good idea for stabilizing our testing and not masking other failures.

PR #1033, opened 2026-09-06
introduces standalone Tile API

lukastruemper:
Refactors most of the LocalStorage additions/experiments into a standalone API which can be easily tested.

Documentation: https://github.com/daisytuner/docc/tree/tile-algebra/opt#tile-algebra-reasoning-about-memory-levels


opt/include/sdfg/targets/omp/plugin.h
```diff
@@ -28,6 +53,8 @@ inline void register_omp_plugin() {
 
     passes::scheduler::SchedulerRegistry::instance()
         .register_loop_scheduler<passes::scheduler::OMPScheduler>(passes::scheduler::OMPScheduler::target());
+
+    tiles::TileTargetRegistry::instance().register_target(ScheduleType_OMP::value(), std::make_shared<OMPTileTarget>());
```
ramonwirsch: We want to reduce global state. Use DoccContext whenever possible instead of introducing new global state that needs to be moved eventually

opt/include/sdfg/tiles/analysis/tile_analysis.h
```diff
@@ -0,0 +1,82 @@
...
+    /// Classify @p container's accesses within @p loop via the pointer analyzers.
+    /// `aliased` is set when the pointer escapes, is overwritten/swapped, or is
+    /// passed to a library node that may capture it — any of which lets the
+    /// container's memory be reached outside the memlets a transformation rewrites.
+    static AccessSummary
+    summarize(const StructuredSDFG& sdfg, structured_control_flow::StructuredLoop& loop, const std::string& container);
+};
+
+/// True iff @p group is a real tile whose extents are all compile-time integer
+/// constants (the constant-bounded localization precondition). A null/empty group
```
ramonwirsch, line 63: Does this mean "immediate" (we know the value) or just "compile-time constant" & "must of scalar-integer type" but can still be a symbolic argument?

opt/src/targets/cuda/plugin.cpp
```diff
@@ -15,6 +19,12 @@ void register_cuda_plugin(plugins::Context& context) {
     auto& reduceDispatcherRegistry = context.reduce_dispatcher_registry;
     auto& libNodeSerRegistry = context.library_node_serializer_registry;
 
+    // The tile algebra's view of CUDA: warp width + level/storage mapping, shared
+    // under both the legacy and the offload schedule value.
+    auto cuda_tile_target = std::make_shared<tiles::GPUTileTarget>(CUDA_WARP_SIZE, ScheduleType_CUDA_Offload::value());
+    tiles::TileTargetRegistry::instance().register_target(ScheduleType_CUDA::value(), cuda_tile_target);
```
ramonwirsch: Same here, we want less global state

opt/src/targets/rocm/plugin.cpp
```diff
@@ -14,6 +18,12 @@ void register_rocm_plugin(plugins::Context& context) {
     auto& reduceDispatcherRegistry = context.reduce_dispatcher_registry;
     auto& libNodeSerRegistry = context.library_node_serializer_registry;
 
+    // The tile algebra's view of ROCm: 64-wide wavefront + level/storage mapping,
+    // shared under both the legacy and the offload schedule value.
+    auto rocm_tile_target = std::make_shared<tiles::GPUTileTarget>(ROCM_WARP_SIZE, ScheduleType_ROCM_Offload::value());
+    tiles::TileTargetRegistry::instance().register_target(ScheduleType_ROCM::value(), rocm_tile_target);
```
ramonwirsch: Same here: less global state

opt/src/tiles/tile.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+AxisSchedule::AxisSchedule(
+    Level level, Space space, bool has_scratchpad, unsigned spatial_axis, symbolic::Integer parallel_size, bool needs_sync
+)
+    : level_(level), space_(space), has_scratchpad_(has_scratchpad), spatial_axis_(spatial_axis),
+      parallel_size_(std::move(parallel_size)), needs_sync_(needs_sync) {}
+
+std::optional<AxisSchedule> AxisSchedule::classify(const structured_control_flow::ScheduleType& sched) {
+    // The target that owns this schedule value supplies the classification.
+    if (auto* target = TileTargetRegistry::instance().get(sched.value())) {
```
ramonwirsch: Impossible to do without global state

opt/src/tiles/analysis/tile_analysis.cpp
```diff
@@ -0,0 +1,285 @@
...
+    ContainerAccessPolicy policy;
+    policy.container = container;
+    ContainerAccessVisitor visitor(sdfg, policy);
+    visitor.visit(loop.root()); // walks the loop body only
+    return AccessSummary{policy.reads, policy.writes, policy.aliased};
+}
+
+void TileAnalysis::run(analysis::AnalysisManager& analysis_manager) {
+    tiles_.clear();
+    auto& mla = analysis_manager.get<analysis::MemoryLayoutAnalysis>();
```
ramonwirsch: This is a very expensive, eager analysis. So even if only needed on a small portion, it will incur compile-time scaling with total SDFG-size.
To save compile-time, we need to transition to a much more local model for expensive analysis.

Can we prepare TileAnalysis for this, so that it could benefit once an analysis like MemoryLayoutAnalysis has been migrated?

opt/src/tiles/tile.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+#include <utility>
+
+#include "sdfg/structured_control_flow/control_flow_node.h"
+#include "sdfg/tiles/tile_target_registry.h"
+
+namespace sdfg {
+namespace tiles {
+
+Space default_space(Level level) {
```
ramonwirsch: Why do we need such a default definition that seems highly GPU-inspired

opt/include/sdfg/tiles/analysis/tile_analysis.h
lukastruemper: No, it needs to be a constant for now, because we need to check the sizes against a sane upper limit

PR #1034, opened 2026-09-07
adds separate einsum namespace

lukastruemper:
Moves einsum sources into its own namespace


Moehre2 (approved):
Documentation could be a little bit more precise... But good for now (I can write a more detailed one when I touch them the next time)

PR #1045, opened 2026-09-10
Fixes for Qwen

Moehre2:
- Fixed bug in DeadCFGElimination that lead to a race condition in the generated code: When trivial loops were removed, access nodes to the indices were not replaced by constant nodes
- Fixed bug in LoopFusionByDomain that deleted blocks because the "perfectly nested" flag was incorrectly set to true after fusing loops that resulted in a non-perfectly nested loop nest
- Added a broadcast functionality on TensorLayout/Tensor that broadcasts the tensor to the provided reference shape by adapting the shape and setting the strides accordingly
- Adapted some elementwise GraphModuleParser's in the PyTorch frontend to automatically try to broadcast
- Fixed a bug in the PyTorch frontend where a constant node with "True"/"False" were generated which is not valid C/C++ code
- Added a unit test for creating attention masks
- Fixed a bug in LoopFusionByAccess that lead to the usage of the wrong index in an access node
- Fixed a bug in LoopFusionByAccess that lead to reading and writing a container twice instead of once
- Improved ArangeNode by adding flop values, adapting the expansion to prevent multiplication of step if step is one and addition of start if start is zero, and bumping test coverage


sdfg/src/passes/structured_control_flow/dead_cfg_elimination.cpp
```diff
@@ -43,6 +52,63 @@ bool DeadCFGElimination::is_trivial(structured_control_flow::Map* loop) {
     return symbolic::eq(trip_count, symbolic::one());
 }
 
+void DeadCFGElimination::
+    update_loop_indvar_accesses(builder::StructuredSDFGBuilder& builder, structured_control_flow::Map* loop) {
```
lukastruemper, line 56: check out "for_each_block" lambda, otherwise we need to maintain this later

lukastruemper (changes requested):
Broadcast logic is sound. DeadCFG is correct, but check out the helper functions. LoopFusion @ramonwirsch needs to check

lukastruemper (approved):
LGTM

PR #1050, opened 2026-09-14
Matrix cores

ramonwirsch:
Many different changes in order to get MMA Matrix core handling for GPUs in.

 * Unified GPUOffloadMap & Reduce Dispatchers in GPUOffloadBaseDispatcher with GPUOffloadDispatcherStrategy handlign the target specific parts for both.
 + we will now write out a header for each kernel, so we can dynamically discover delcarations and new includes inside kernels as well
! deprecating LibraryNode.implementation_type() mutable ref for .set_implementation_type() to be prepared for changing the implementation type
! deprecating LibraryNodeDispatcherRegistry.register_library_node_dispatcher(std::string, std::function) for a version that gets LibraryNodeCode and ImplementationType separately, also preparing for switching out ImplType without needing to change the calling code
+ LibraryNodeExpansionPass & Visitor are now templated to support more than 1 expander per pass in the future
+ SrcFileCompilerBuilder.set_compile_order() to constrain that certain files (gpu kernels) can only be compiled after all possible header snippets have been generated. Prevents crashes on many-core machines where the headers did not yet exist.
 + SrcFileCompilerBuilder now is fine with not setting main-ext for snippet redirects. Not needed.

 + rocm-specific test that tries to Transform a sliced matmul into using MMA cores for MI210 and RX9070 XT
 * extended MatMulNode to support a sliced (explicit) result TensorLayout.
 ! currently using the offset property of TensorLayouts to for the inner-tensor offsets, even though in ptr-input cases, it really should be folded into the inputs instead (but we currently do not have simple pointer-math on inputs edges into lib-nodes)
 + Started on GpuArch and GpuMmaSupport generic modelling, implemented for Cuda and Rocm that provide all the needed info in order to find valid MMA combinations. But currently not integrated with target, impl-type, schedule type and just an explicit argument into the expansion.

MMA Expansion: require sliced Matmul be exact multiple of MMA block size (currently strictly 16x16x16 for ROCM), choosing either 1x1, (1-2)x(1-2) or (4|8)x(1-4) wavefronts per block, all with 16x16x16  blocks, but correct wavefront size for the supported architectures.
ROCM MMA Transform: targeted into the inner sliced Matmul above, inside some outer-loop with ScheduleType ROCM and targeting the grid dimensions.

Currently, uses arch specific impl-types (ROCM_MMA_GFX90A, ROCM_MMA_GFX1201), because we don't have a good way prepared to carry the exact architecture information from the expander to the dispatcher / library node.
Longterm goal is to make ImplTypes customizable and also likely MMA specific libNodes that can carry that information directly and provide more tuning control via arguments.


lukastruemper (approved):
LGTM

